PR #5, opened 2019-09-06
Drop HAVE_GETFSSTAT in Check_HR_FileSys_AutoFs

slashdd:
GETFSSTAT doesn't exist in Linux, and the preprocessor directive
looks unnecessary in this function.

Reference:
https://sourceforge.net/p/net-snmp/patches/1350/

gcc -E without that commit:

int
Check_HR_FileSys_AutoFs(void)
{
        return 1;

    return 0;
}

gcc -E with that commit:

int
Check_HR_FileSys_AutoFs(void)
{
    if (HRFS_entry->HRFS_type != NULL &&

        !strcmp(HRFS_entry->HRFS_type, "autofs")

        )
        return 1;

    return 0;
}

### DIFF
--- a/agent/mibgroup/host/hr_filesys.c
+++ b/agent/mibgroup/host/hr_filesys.c
@@ -841,7 +841,6 @@ Check_HR_FileSys_NFS (void)
 int
 Check_HR_FileSys_AutoFs(void)
 {
-#if HAVE_GETFSSTAT
     if (HRFS_entry->HRFS_type != NULL && 
 #if defined(MNTTYPE_AUTOFS)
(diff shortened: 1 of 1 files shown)


bvanassche:
The output of git grep -nH 'HRFS_type = ' tells me that HRFS_type is only set on AIX systems but not on Linux systems. Do you agree with this conclusion?

slashdd:
Disregard this PR please. Thanks

slashdd:
Fix via https://github.com/net-snmp/net-snmp/commit/bcb1a6b8afc4

Thanks !

PR #7, opened 2019-09-10
linux: IF-MIB Counters (HwIfInOctets, HwIfOutOctets, etc.) became 32-bit counters

stannous:
Recently, patch 64f4ead (libsnmp: Only include <inttypes.h> when necessary)
effectively removed inttypes.h from types.h and added it back for most files
except for

if-mib/data_access/interface_linux.c

This turned what were 64 bit counters (SCNuMAX is no longer defined) into
32 bit counters.

This patch simply adds inttypes.h back to interface_linux.c
correcting the counters.


bvanassche:
Github tells me that this pull request includes 307 commits. Did you perhaps merge the master branch into the v5.8 branch? Anyway, I have cherry-picked your patch onto the v5.8 and master branches. Thanks for the patch!

PR #13, opened 2019-09-25
Perl: Allow informs to be sent with SNMP v2(c).

pmorch:
Before informs required SNMP v3. But informs actually work just fine
with v2c too too.

### DIFF
--- a/perl/SNMP/SNMP.pm
+++ b/perl/SNMP/SNMP.pm
@@ -1234,7 +1234,7 @@ sub inform {
    my $trap_oid = $param{oid} || $param{trapoid};
    my $uptime = $param{uptime} || SNMP::_sys_uptime();
 
-   if($this->{Version} eq '3') {
+   if ($this->{Version} =~ '^[23]') {
      @res = SNMP::_inform($this, $uptime, $trap_oid, $varbind_list_ref, $cb);
    } else {
      warn("error:inform: This version doesn't support the command\n");


hardaker:
The version number can also be '2c' and the regexp should take care of that case too.  I'd suggest also using //s, which are more common for perl.  Maybe /(2c?|3)/ ?

pmorch:
Wow, you're right about `//` instead of `''`. Sorry about that. I'll fix that in a next commit.

But I don't understand what you mean with:

> The version number can also be '2c' and the regexp should take care of that case too

What doesn't work about the logic?:

```
» perl -w -e '                                            
foreach my $version ("1", "2", "2c", "3") {
     printf "%-2s: %s\n",                                         
          $version,                                     
          ($version =~ "^[23]") ?    
               "match": "no match";
} 
'
1 : no match
2 : match
2c: match
3 : match
```

pmorch:
Actually there is another (small) problem with the logic around line 1199:

```
if ($this->{Version} eq '1') {
    # trap code versions 1
} elsif ($this->{Version} =~ '^[23]') {
    # trap code for later versions
}
```
What about other/unknown versions? They silently don't do anything. I suggest that should be either:
```
if ($this->{Version} eq '1') {
    # trap code versions 1
} elsif ($this->{Version} =~ '^[23]') {
    # trap code for later versions
} else {
    die "Unknown version: " . $this->{Version};
}
  
```
or simply:
```
if ($this->{Version} eq '1') {
    # trap code versions 1
} else {
    # trap code for later versions
} 
```

If we go with the latter, then, following the same logic, the patch in this PR should be logically inverted and just check for `$this->{Version} eq '1'`.

```
if ($this->{Version} eq '1') {
  warn("error:inform: This version doesn't support the command\n");
} else {
  @res = SNMP::_inform($this, $uptime, $trap_oid, $varbind_list_ref, $cb);
}
```

pmorch:
But perhaps the path of least resistance and least risk is just replacing `''` with `//` and be done with it.

bvanassche:

Issue #12, opened 2019-09-25
Linux traps with VRF iface are broken unless clientaddr is used


stannous:
(I had incorrectly opened a "patch" for this bug a while ago (#1399).)

Patches 0831ed (main Linux VRF support) and 3ca90c2c (added use of sendto()
instead of sendmsg()) added basic linux vrf support.

But on latest V5-8-patches, any trapsink, trap2sink, or trapsess with a VRF interface

```
trapsess -e 0x80001f88806d7e38395026435d00000000 -v3 -l authPriv -u testuser -a SHA -A testsha1234 -x AES -X testaes1234 11.0.0.1%red 
trap2sink 11.0.0.1@red public  
trapsink 11.0.0.1@red public 
```

does not send a trap out the right interface.

At one point, v1 and v2c traps were working (not sure if I ever got v3 traps working) with 
linux VRFs.

stannous:
I forgot to mention that using a source IP address for traps 
`clientaddr 11.0.0.10@red` in /etc/snmp/snmp.conf **fixes** this problem.
This sort of makes sense.  But I can't find when the breakage occurred.

Having said that, our users do not usually use "clientaddr"  and I think that we were
setting SO_BINDTODEVICE correctly before this behavior changed.

It's possible that when netsnmp_udpbase_sendto_unix() checks for
srcip && srcip->s_addr != INADDR_ANY, it is being too restrictive.
If the socket has SO_BINDTODEVICE set to the correct iface, we should
be using sendto() instead of sendmsg().  That could explain why this no
longer works.  I'll check this.

--Sam

PR #18, opened 2019-09-26
Linux VRF iface on Trap sink IP address not being set

stannous:
For Linux VRF support in Traps, this patch fixes an issue
(#12) where configuring and iface on the sink IP address
was not being bound (via SO_BINDTODEVICE) to the socket.
This was broken for versions 1, 2c, and 3.

While a clientaddr setting in snmp.conf could fix this,
a user would have to find and set a source IP address
in the VRF to be used. This patch does not require
setting a clientaddr but simply binds the iface
provided in the sink IP address (target) to the
session socket that will be used.

### DIFF
--- a/snmplib/transports/snmpUDPIPv4BaseDomain.c
+++ b/snmplib/transports/snmpUDPIPv4BaseDomain.c
@@ -301,6 +301,17 @@ netsnmp_udpipv4base_transport_with_source(const struct netsnmp_ep *ep,
     if (NULL == bind_addr)
         return t;
 
+    /* for Linux VRF Traps we try to bind the iface if clientaddr is not set */
+    if (ep && ep->iface && t->sock) {
+        rc =  netsnmp_bindtodevice(t->sock, ep->iface);
+        if (rc)
+            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Could not bind socket %d bound to %s\n",
+                        t->sock, ep->iface));
+        else
+            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Bound socket %d bound to %s\n",
+                        t->sock, ep->iface));
+    }
+
     rc = netsnmp_udpipv4base_transport_bind(t, bind_addr, flags);
     if (rc) {
         netsnmp_transport_free(t);
--- a/snmplib/transports/snmpUDPIPv4BaseDomain.c
+++ b/snmplib/transports/snmpUDPIPv4BaseDomain.c
@@ -302,13 +302,13 @@ netsnmp_udpipv4base_transport_with_source(const struct netsnmp_ep *ep,
         return t;
 
     /* for Linux VRF Traps we try to bind the iface if clientaddr is not set */
-    if (ep && ep->iface && t->sock) {
+    if (ep && ep->iface) {
         rc =  netsnmp_bindtodevice(t->sock, ep->iface);
         if (rc)
-            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Could not bind socket %d bound to %s\n",
+            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Could not bind socket %d to %s\n",
                         t->sock, ep->iface));
         else
-            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Bound socket %d bound to %s\n",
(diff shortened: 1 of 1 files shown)


snmplib/transports/snmpUDPIPv4BaseDomain.c
```diff
@@ -301,6 +301,17 @@ netsnmp_udpipv4base_transport_with_source(const struct netsnmp_ep *ep,
     if (NULL == bind_addr)
         return t;
 
+    /* for Linux VRF Traps we try to bind the iface if clientaddr is not set */
```
bvanassche: This line is only reached if bind_addr != NULL. If clientaddr has not been set then src_addr == NULL. So this comment is not clear to me.

snmplib/transports/snmpUDPIPv4BaseDomain.c
```diff
@@ -301,6 +301,17 @@ netsnmp_udpipv4base_transport_with_source(const struct netsnmp_ep *ep,
...
+    if (ep && ep->iface && t->sock) {
+        rc =  netsnmp_bindtodevice(t->sock, ep->iface);
```
bvanassche: The t->sock test looks wrong to me since 0 is a valid socket descriptor. Additionally, I think if this if-test is reached that it is already guaranteed that t->sock >= 0 and hence that it is not necessary to test t->sock.

Shouldn't netsnmp_bindtodevice() only be called if local != 0, in other words, if a socket is being created for sending data? I think local == 0 means that a listening socket is being created.

snmplib/transports/snmpUDPIPv4BaseDomain.c
```diff
@@ -301,6 +301,17 @@ netsnmp_udpipv4base_transport_with_source(const struct netsnmp_ep *ep,
...
+        if (rc)
+            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Could not bind socket %d bound to %s\n",
```
bvanassche: "bound" looks superfluous here?

snmplib/transports/snmpUDPIPv4BaseDomain.c
```diff
@@ -301,6 +301,17 @@ netsnmp_udpipv4base_transport_with_source(const struct netsnmp_ep *ep,
...
+                        t->sock, ep->iface));
+        else
+            DEBUGMSGTL(("netsnmp_udpbase", "VRF: Bound socket %d bound to %s\n",
```
bvanassche: Two times "bound" in one message?

snmplib/transports/snmpUDPIPv4BaseDomain.c
fenner: The Linux VRF API is a little weird - they overload the "bind to device" to mean "live inside the VRF that this interface is configured to be in".  So it makes sense to bind a listen socket to a VRF, even if in the wider view it doesn't make sense to bind a listening socket to a specific interface.

snmplib/transports/snmpUDPIPv4BaseDomain.c
stannous: we do have a bind_addr (sink) but we also have an iface we need to bind on the socket.  Otherwise, at least for linux, we will not send the trap out the right interface.

snmplib/transports/snmpUDPIPv4BaseDomain.c
stannous: right.  you can think of the VRF as a master interface (like a bridge)...which internally, I think it is.

snmplib/transports/snmpUDPIPv4BaseDomain.c
stannous: ack...thanks.

snmplib/transports/snmpUDPIPv4BaseDomain.c
stannous: yes, thanks.

bvanassche:
The proper procedure for making changes to a pull request that includes a single patch is as follows:
- Make the changes locally, review and test these.
- Amend the already published patch with git commit -a --amend
- Modify the pull request with git push ${repository_name} +${branch_name}

Anyway, I have combined the two patches into a single patch and applied the result on the v5.8 and master branches. Please take a look.

PR #24, opened 2019-10-09
Reduce CPU load for IF-MIB on Linux servers with large number of interfaces

stannous:
Reduce CPU load for IF-MIB on Linux servers with large number of interfaces

This patch introduces a new configution option to limit the number
of interfaces that the IF-MIB will process.  On Linux servers with a
large number of interfaces (ppp, dummy, bridge, etc.), the IF-MIB
timer based stats collector can take a large amount of CPU processing.

The new config option "include_ifmib_iface_prefix" takes a space
separated string of ifname prefixes ("eth lo bridg") and will only
include interfaces with these prefixes in the IF-MIB tables.

If this config option is not present (the default), all interfaces
will continue to be processed in the IF-MIB.

This patch is largely based on work by Collet Thibaut from 

https://sourceforge.net/p/net-snmp/patches/1352/ 

### DIFF
--- a/agent/mibgroup/if-mib/data_access/interface.c
+++ b/agent/mibgroup/if-mib/data_access/interface.c
@@ -32,7 +32,7 @@ netsnmp_feature_require(interface_arch_set_admin_status);
 static netsnmp_conf_if_list *conf_list = NULL;
 static int need_wrap_check = -1;
 static int _access_interface_init = 0;
-static netsnmp_include_if_list *include_list = NULL;
+static netsnmp_include_if_list *include_list;
 
 /*
  * local static prototypes
@@ -49,6 +49,7 @@ static void _free_interface_config(void);
 static void _parse_include_if_config(const char *token, char *cptr);
 static void _free_include_if_config(void);
 
+SNMPCallback _prune_include_if_list;
 
 /**
  * initialization
@@ -59,13 +60,20 @@ init_interface(void)
     snmpd_register_config_handler("interface", _parse_interface_config,
                                   _free_interface_config,
                                   "name type speed");
-    snmpd_register_config_handler("include_ifmib_iface_prefix",
-                                   _parse_include_if_config,
-                                   _free_include_if_config,
-                                   "IF-MIB iface names included");
-
 }
 
+void
+netsnmp_include_interface_init(void)
+{
+    snmpd_register_config_handler("include_ifmib_iface_prefix",
+                                  _parse_include_if_config,
+                                  _free_include_if_config,
+                                  "IF-MIB iface names included");
+
+    snmp_register_callback(SNMP_CALLBACK_LIBRARY,
+                           SNMP_CALLBACK_POST_READ_CONFIG,
+                           _prune_include_if_list, NULL);
+}
 
 void
 netsnmp_access_interface_init(void)
@@ -794,19 +802,19 @@ netsnmp_access_interface_entry_overrides(netsnmp_interface_entry *entry)
 /*
  * include_ifmib_iface_prefix config token
  *
- *     If there is an iface name match, we return 1 for True
+ *     If there is an iface prefix name match, we return 1 for True
  *     otherwise return 0 for False.
  *     If there are no ifaces defined at all, return 1 so that the
  *     default behavior is to include all ifaces (include everything).
- *     (Note: including at least one iface name means we will only include
- *     those iface names and exclude all others.)
+ *     (Note: including at least one iface prefix means we will only include
+ *     those iface names that match the prefix and exclude all others.)
  *
  */
 int netsnmp_access_interface_include(const char * name)
 {
-    netsnmp_include_if_list * if_ptr;
+    netsnmp_include_if_list *if_ptr;
 
-    netsnmp_assert(1 == _access_interface_init);
+    /*netsnmp_assert(_access_interface_init == 1);*/
     if (NULL == name)
         return 1;
 
@@ -818,7 +826,7 @@ int netsnmp_access_interface_include(const char * name)
         return 1;
 
     for (if_ptr = include_list; if_ptr; if_ptr = if_ptr->next)
-        if (memcmp(name, if_ptr->name, strlen(if_ptr->name)) == 0)
+        if (strncmp(name, if_ptr->name, strlen(if_ptr->name)) == 0)
             return 1;
 
     return 0;
@@ -900,6 +908,25 @@ _free_interface_config(void)
     conf_list = NULL;
 }
 
+int
+_prune_include_if_list(int majorID, int minorID,
+                       void *serverargs, void *clientarg)
+{
+    netsnmp_container * ifcontainer;
+
+    _access_interface_init = 1;
+    netsnmp_arch_interface_init();
+
+    /*
+     * load once to set up ifIndexes
+     */
+    ifcontainer = netsnmp_access_interface_container_load(NULL, 0);
+    if(NULL != ifcontainer)
+        netsnmp_access_interface_container_free(ifcontainer, 0);
+
+    return SNMPERR_SUCCESS;
+}
+
 /**---------------------------------------------------------------------*/
 /*
  * include interface config token
@@ -910,24 +937,23 @@ static void
 _parse_include_if_config(const char *token, char *cptr)
 {
     netsnmp_include_if_list *if_ptr, *if_new;
-    char                   *name, *st;
+    char                    *name, *st;
 
     name = strtok_r(cptr, " \t", &st);
     if (!name) {
         config_perror("Missing NAME parameter");
         return;
     }
-    if_ptr = include_list;
-    while (if_ptr)
-        if (strcmp(if_ptr->name, name))
-            if_ptr = if_ptr->next;
-        else
-            break;
-    if (if_ptr) {
-        config_pwarn("Duplicate include interface specification");
-        return;
-    }
+
+    /* check for duplicate prefix configuration */
     while (name != NULL) {
+        for (if_ptr = include_list; if_ptr; if_ptr = if_ptr->next) {
+            if (strncmp(name, if_ptr->name, strlen(if_ptr->name)) == 0) {
+                config_pwarn("Duplicate include interface prefix specification");
+                return;
+            }
+        }
+        /* now save the prefixes */
         if_new = SNMP_MALLOC_TYPEDEF(netsnmp_include_if_list);
         if (!if_new) {
             config_perror("Out of memory");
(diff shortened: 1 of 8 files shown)


agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -32,6 +32,7 @@ netsnmp_feature_require(interface_arch_set_admin_status);
 static netsnmp_conf_if_list *conf_list = NULL;
 static int need_wrap_check = -1;
 static int _access_interface_init = 0;
+static netsnmp_include_if_list *include_list = NULL;
```
bvanassche: Please leave out the " = NULL" initializer - it is superfluous. See also https://en.wikipedia.org/wiki/.bss.

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -782,6 +790,40 @@ netsnmp_access_interface_entry_overrides(netsnmp_interface_entry *entry)
...
+/**---------------------------------------------------------------------*/
+/*
+ * include_ifmib_iface_prefix config token
+ *
+ *     If there is an iface name match, we return 1 for True
+ *     otherwise return 0 for False.
+ *     If there are no ifaces defined at all, return 1 so that the
+ *     default behavior is to include all ifaces (include everything).
+ *     (Note: including at least one iface name means we will only include
+ *     those iface names and exclude all others.)
```
bvanassche: Please mention in this comment that a prefix match is performed.

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -782,6 +790,40 @@ netsnmp_access_interface_entry_overrides(netsnmp_interface_entry *entry)
...
+ *
+ */
+int netsnmp_access_interface_include(const char * name)
+{
+    netsnmp_include_if_list * if_ptr;
+
+    netsnmp_assert(1 == _access_interface_init);
```
bvanassche: Please put the constant at the right of "==". All C compilers today issue a warning when using an assignment in an if-condition.

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -782,6 +790,40 @@ netsnmp_access_interface_entry_overrides(netsnmp_interface_entry *entry)
...
+ *     otherwise return 0 for False.
+ *     If there are no ifaces defined at all, return 1 so that the
+ *     default behavior is to include all ifaces (include everything).
+ *     (Note: including at least one iface name means we will only include
+ *     those iface names and exclude all others.)
+ *
+ */
+int netsnmp_access_interface_include(const char * name)
+{
+    netsnmp_include_if_list * if_ptr;
```
bvanassche: I think the Net-SNMP coding style requires only a space in front of "*" but not after. See also the CodingStyle file.

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -782,6 +790,40 @@ netsnmp_access_interface_entry_overrides(netsnmp_interface_entry *entry)
...
+
+    if (include_list == NULL)
+        /*
+         * If include_ifmib_iface_prefix was not configured, we should include
+         * all interfaces (which is the default).
+         */
+        return 1;
+
+    for (if_ptr = include_list; if_ptr; if_ptr = if_ptr->next)
+        if (memcmp(name, if_ptr->name, strlen(if_ptr->name)) == 0)
```
bvanassche: Can this memcmp() call be changed into a strncmp() call? I'm concerned that memcmp() could read past the end of 'name' and that that could trigger a Valgrind complaint.

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -857,3 +899,61 @@ _free_interface_config(void)
...
+    if_ptr = include_list;
+    while (if_ptr)
+        if (strcmp(if_ptr->name, name))
+            if_ptr = if_ptr->next;
+        else
+            break;
+    if (if_ptr) {
+        config_pwarn("Duplicate include interface specification");
+        return;
+    }
```
bvanassche: So this loop only checks whether the first interface name is unique? Shouldn't this check be moved inside the while(name) loop?

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -857,3 +899,61 @@ _free_interface_config(void)
...
+}
+
+static void
+_free_include_if_config(void)
+{
+    netsnmp_include_if_list   *if_ptr = include_list, *if_next;
+    while (if_ptr) {
+        if_next = if_ptr->next;
+        free(NETSNMP_REMOVE_CONST(char *, if_ptr->name));
```
bvanassche: Is it really useful to declare the name member 'const'? It would make me happy if the above cast would disappear.

agent/mibgroup/if-mib/data_access/interface.c
stannous: Will do. Thanks.

agent/mibgroup/if-mib/data_access/interface.c
stannous: ok.

agent/mibgroup/if-mib/data_access/interface.c
stannous: ok.

agent/mibgroup/if-mib/data_access/interface.c
stannous: ack.

agent/mibgroup/if-mib/data_access/interface.c
stannous: Ack.  I'll test a change without const.

agent/mibgroup/if-mib/data_access/interface.c
stannous: You're right.  I just took Collet original patch without checking this.  I'll test a fix for this.

agent/mibgroup/if-mib/data_access/interface.c
stannous: Yes, I can change it.

agent/mibgroup/if-mib/data_access/interface.c
stannous: So when I change this to a single loop, the IF-MIB interface filtering (index_list) doesn't
get initialized until after init_mib_modules() so that for the first 5 minutes or so after
snmpd is started, I see all the interfaces (over 1000).  After about 5 minutes, my interface
list is down to about 54 (which I configured with "include_ifmib_iface_prefix". 

Do you know of a way to get the "include" list to take effect right when snmpd is started?
It looks like the IF-MIB initialization (with a timer) is started before the new configuration
is parsed and my index_list is created.  I think this is caused this.

```
+    /* check for duplicate prefix configuration */
+    while (name != NULL) {
+        for (if_ptr = include_list; if_ptr; if_ptr = if_ptr->next) {
+            if (strncmp(name, if_ptr->name, strlen(if_ptr->name)) == 0) {
+                config_pwarn("Duplicate include interface prefix specification");
+                return;
+            }
+        }
+        /* now save the prefixes */
+        if_new = SNMP_MALLOC_TYPEDEF(netsnmp_include_if_list);
+        if (!if_new) {
+            config_perror("Out of memory");
+            return;
+        }
+        if_new->name = strdup(name);
+        if (!if_new->name) {
+            config_perror("Out of memory");
+            free(if_new);
+            return;
+        }
+        if_new->next = include_list;
+        include_list = if_new;
+        name = strtok_r(NULL, " \t", &st);
+    }
+}

```

bvanassche: How about moving snmpd_register_config_handler("include_ifmib_iface_prefix", ...) from init_interface() into init_mib_modules(), just before the netsnmp_access_interface_init() call?

stannous: I tried moving the call to snmpd_register_config_handler for the "include_ifmib_iface_prefix"
to the beginning of init_mib_modules:

```
@@ -65,6 +65,8 @@ init_mib_modules(void)
 {
     static int once = 0;
 
+    netsnmp_include_interface_init();
+
```
It still takes 5 minutes for my interface count to drop from 1056 to 56.
I would venture to guess that the initialization is happening without
my check for index_list of prefixes.  I'm probably just missing a call
for the check: 

```
+ if (!netsnmp_access_interface_include(ifrp->ifr_name)) 
+            continue; 
+ 

```

bvanassche:
How about registering a POST_READ_CONFIG callback and gathering network interface information for the first time when that callback is called? From init_mteObjects():
~~~
    snmp_register_callback(SNMP_CALLBACK_LIBRARY, 
                           SNMP_CALLBACK_POST_READ_CONFIG,
                           _init_default_mteObject_lists, NULL);
~~~

stannous:
From what I can tell, the file agent/mibgroup/if-mib/ifTable/ifTable_interface.c
is not editable.  

So even though I can register a callback for SNMP_CALLBACK_POST_READ_CONFIG,
I'm not sure how to limit (or prune) the interface list that is somehow generated
in ifTable_interface.c.  There's some initialization magic (m2c?) that has to happen
to prevent *all* interfaces from being included until the 300 second cache timer pops.

Is there an easy way to bring up snmpd and remove (or change) the interfaces
included in the IF-MIB?  How would I do this from my post-read callback function?

Any hints would be appreciated.

stannous:
I moved the config handler registration to agent/mib_modules.c:init_mib_modules and
added a new function for SNMP_CALLBACK_POST_READ_CONFIG.  There, 
since I have my filter list (include_list) of interface prefixes, I attempted to initialize
the interface container (in _prune_include_if_list()).  But there's still something wrong
with this approach since the iF-MIB seems to ignore my filtering attempt and the full
list of interfaces is still included in this container.   So even though I get a breakpoint
in netsnmp_arch_interface_container_load() and do my filtering correctly, the initial
container is somehow initialized with all the interfaces.

So in looking at  init_mteObjects(), I do have my filter list of interface prefixe names
to include in my _prune_include_if_list() but I'm not sure what to do with this.
Initializing the IF-MIB container doesn't seem to work.  What I need to do is prevent the
"normal" IF-MIB initialization or at least make sure it uses my prefix filter list.

bvanassche:
Please have a look at the following patch and see whether network interface filtering works properly with that patch: https://github.com/bvanassche/net-snmp/commit/76ad71101ce0a4db50f5ab9dd025b13874456b55

stannous:
bvanassche@76ad711 works fine.  Thanks!

bvanassche:
Closing this pull request since the patch mentioned in the previous comment has been checked in.

PR #31, opened 2019-10-18
libsnmp: expose snmp_build and snmp_parse

Enlik:


Enlik:
For what it's worth,
as for now my main concern is about static libraries but I also checked the shared one, and even without the patch, this function is already exposed (at least on a Linux build), in libnetsnmp.so:

```
000000000003a6f0 T snmp_build
```

but still I believe it should be marked with `NETSNMP_IMPORT`. Hope it helps with the review.

bvanassche:
I have modified snmp_parse() such that it accepts a single session pointer instead of two before I checked in your patch. Please have a look at the v5.8 branch or the master branch.

Enlik:
Thank you, I have a remark though.

In my case, net-snmp functions are called without a session list pointer, e.g.: `snmp_sess_init(session)`, `snmp_error(session, …)`, `snmp_build(…, session)`

and, in case of snmp_parse: `snmp_parse(NULL, session, …)`.

And it works; after your change, the code would be forced to create a session list. What do you think?

One one hand:
In `netsnmp_sess_api.3` several functions that work like this are documented, so it's a supported usage.

On the other hand:
when it comes to `snmp_build`: it handles `NULL` as `sessp / slp` gracefully, with a slight problem in case of SNMPv3 where it descends to `_snmp_parse -> snmp_sess_send -> snmp_sess_async_send` where this is done:

```
    if (sessp == NULL) {
        snmp_errno = SNMPERR_BAD_SESSION;       /*MTCRITICAL_RESOURCE */
        return (0); 
    }
```

(it's in the 5.8 branch, 8b957a683). Not sure if it's a "problem" or one of supported cases. If the latter, being allowed not to create a "session list" would be useful.

bvanassche:
Are you aware that it is easy to convert a struct session_list pointer into a netsnmp_session pointer and vice versa? See also snmp_sess_pointer() and the session member of struct session_list.

Enlik:
From what I can see, `snmp_sess_pointer` requires that the session has been created in the first place (e.g. `snmp_open`), or added "manually" to Sessions with something like `snmp_session_insert` after creating it. `snmp_sess_open` does quite a few things that hasn't been necessary (the application needs just snmp_build and snmp_parse).

Sorry for making the PR into kind of a support thread, but I hope you advice after this API change. :)

bvanassche:
The netsnmp_session argument of snmp_parse() has been restored. Please have another look.

Enlik:
From my perspective everything is fine. Thank you! Just to make sure, can I/you close the PR or are you planning to do something that will affect the functionality?

bvanassche:
Thanks for the feedback. I was waiting for your feedback to close this pull request.

PR #36, opened 2019-11-13
Add new config option to limit max number of interfaces for IF-MIB

stannous:
This patch introduces a new config option "ifmib_max_num_ifaces NUM"
This patch sets the maximum number of interfaces included in IF-MIB
data collection.  For servers with a large number of interfaces
(ppp, dummy, bridge, etc) the IF-MIB processing will take a large
chunk of CPU for ioctl calls (on Linux). Setting a reasonable
maximum number of ifaces for the CPU used will reduce the CPU
load for IF-MIB processing.  For example, configuring
"ifmib_max_num_ifaces 500" will include only the first 500
interfaces based on ifindex and ignore all others for IF-MIB
processing.  If missing, the default is to not limit the maximum
number of interfaces (current default).

### DIFF
--- a/agent/mibgroup/if-mib/data_access/interface.c
+++ b/agent/mibgroup/if-mib/data_access/interface.c
@@ -946,7 +946,7 @@ _free_interface_config(void)
 static void
 _parse_ifmib_max_num_ifaces(const char *token, char *cptr)
 {
-    long long_val;
+    int temp_max;
     char *name, *st;
 
     errno = 0;
@@ -955,13 +955,12 @@ _parse_ifmib_max_num_ifaces(const char *token, char *cptr)
         config_perror("Missing NUMBER parameter");
         return;
     }
-    long_val = strtol(cptr, NULL, 10);
-    if (errno) {
-           config_perror("Error converting parameter");
-           return;
+    if (sscanf(cptr, "%d", &temp_max) != 1) {
+        config_perror("Error converting parameter");
+        return;
     }
 
-    ifmib_max_num_ifaces = (int) long_val;
+    ifmib_max_num_ifaces = temp_max;
 }
 
 
--- a/agent/mibgroup/if-mib/data_access/interface.c
+++ b/agent/mibgroup/if-mib/data_access/interface.c
@@ -6,6 +6,7 @@
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/net-snmp-features.h>
 #include <net-snmp/net-snmp-includes.h>
+#include <errno.h>
 
 #include <net-snmp/agent/net-snmp-agent-includes.h>
 #include <net-snmp/library/snmp_enum.h>
@@ -33,6 +34,7 @@ static netsnmp_conf_if_list *conf_list = NULL;
 static int need_wrap_check = -1;
 static int _access_interface_init = 0;
 static netsnmp_include_if_list *include_list;
+static int ifmib_max_num_ifaces = 0;
 
 /*
  * local static prototypes
@@ -45,6 +47,7 @@ static void _access_interface_entry_release(netsnmp_interface_entry * entry,
 #endif
 static void _access_interface_entry_save_name(const char *name, oid index);
 static void _parse_interface_config(const char *token, char *cptr);
+static void _parse_ifmib_max_num_ifaces(const char *token, char *cptr);
 static void _free_interface_config(void);
 static void _parse_include_if_config(const char *token, char *cptr);
 static void _free_include_if_config(void);
@@ -69,6 +72,11 @@ init_interface(void)
     snmpd_register_config_handler("interface", _parse_interface_config,
                                   _free_interface_config,
                                   "name type speed");
+
+    snmpd_register_config_handler("ifmib_max_num_ifaces",
+                                  _parse_ifmib_max_num_ifaces,
+                                  NULL,
+                                  "IF-MIB MAX Number of ifaces");
 
     snmpd_register_config_handler("include_ifmib_iface_prefix",
                                   _parse_include_if_config,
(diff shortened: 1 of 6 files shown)


fenner:
Hi Sam, I appreciate the amount of work you're putting into addressing this unfortunate limitation.  I've run into it on Ubiquiti gear when creating a large number of interfaces, and it just meant that we had to turn off snmpd, so allowing *something* to work is a blessing.

What's the fundamental reason that cpu usage is too high?  Is it because we poll too often, or maybe because there's a hidden n**2 algorithm at work?  I wonder if we would be better off trying to find the root of the problem.  Maybe only poll counters every 30 or 60 seconds if that's what's causing the load (or create a configuration value for that?)

stannous:
I didn't see any evidence of a hidden n**2 algorithm.  The high CPU seems to be the
sheer number of interfaces (perhaps the open/reads from /proc/net/dev which happens
every 3 seconds (default)).

This patch is simply an attempt to provide an upper limit default setting for users 
to prevent the CPU usage for snmpd (80-100%) for those users who have a need
for a large number of interfaces.

A longer term solution might be to rewrite netsnmp_arch_interface_container_load() for linux
to replace the reading of /proc/net/dev with something more efficient (perhaps sysctl).

bvanassche:
Does ethtool -S produce all the statistics Net-SNMP needs? ethtool source code is available at [https://git.kernel.org/pub/scm/network/ethtool/ethtool.git](https://git.kernel.org/pub/scm/network/ethtool/ethtool.git).

```
$ ethtool -S eno1
NIC statistics:
     rx_packets: 284182
     tx_packets: 175023
     rx_bytes: 322280490
     tx_bytes: 21353950
     rx_broadcast: 16
     tx_broadcast: 2
     rx_multicast: 1291
     tx_multicast: 2231
     rx_errors: 0
     tx_errors: 0
     tx_dropped: 0
     multicast: 1291
     collisions: 0
     rx_length_errors: 0
     rx_over_errors: 0
     rx_crc_errors: 0
     rx_frame_errors: 0
     rx_no_buffer_count: 0
     rx_missed_errors: 0
     tx_aborted_errors: 0
     tx_carrier_errors: 0
     tx_fifo_errors: 0
     tx_heartbeat_errors: 0
     tx_window_errors: 0
     tx_abort_late_coll: 0
     tx_deferred_ok: 0
     tx_single_coll_ok: 0
     tx_multi_coll_ok: 0
     tx_timeout_count: 0
     tx_restart_queue: 0
     rx_long_length_errors: 0
     rx_short_length_errors: 0
     rx_align_errors: 0
     tx_tcp_seg_good: 0
     tx_tcp_seg_failed: 0
     rx_flow_control_xon: 0
     rx_flow_control_xoff: 0
     tx_flow_control_xon: 0
     tx_flow_control_xoff: 0
     rx_csum_offload_good: 282844
     rx_csum_offload_errors: 2
     rx_header_split: 0
     alloc_rx_buff_failed: 0
     tx_smbus: 0
     rx_smbus: 0
     dropped_smbus: 0
     rx_dma_failed: 0
     tx_dma_failed: 0
     rx_hwtstamp_cleared: 0
     uncorr_ecc_errors: 0
     corr_ecc_errors: 0
     tx_hwtstamp_timeouts: 0
     tx_hwtstamp_skipped: 0
```

stannous:
Bart, it looks like netsnmp_arch_interface_container_load() for linux is just looping over all the
interfaces in /proc/net/dev and for each interface, there are ioctl calls to get the hw address,
get ifindex, check if it has an ipv4 address, flags, and MTU (for duplex and speed, we use ethtool
ETHTOOL_GLINKSETTINGS ioctl).  

So for 2000 interfaces, I'd estimate we have at least 10,000 ioctl calls every 3 seconds.
We also parse the data in /proc/net/dev to collect some stats:

    uintmax_t       rec_pkt, rec_oct, rec_err, rec_drop, rec_mcast;                                                                                                                                
    uintmax_t       snd_pkt, snd_oct, snd_err, snd_drop, coll;     

So even if we collect these counters from an ioctl call rather then /proc/net/dev,
I can't see how this is going to help much.  We still have ~10k ioctl calls to make.

Some ideas that may help us: we could make the MTU, speed, mac address calls
less frequently.  Or maybe collect stats from sysfs /sys/class/net/ethXX/statistics/ ?
There may be some other mechanism for linux that we can leverage.  I can keep looking.

bvanassche:
I think the following statement from netsnmp_arch_interface_container_load() parses the /proc/net/dev network statistics:
```
_parse_stats(entry, stats, scan_expected);
```
ETHTOOL_GLINKSETTINGS is a legacy ioctl that only queries link settings. There are newer and much more interesting ioctls, e.g. ETH_SS_STATS and ETHTOOL_GSTATS. From the ethtool soure code:
```
static int do_gnicstats(struct cmd_context *ctx)
{
	return do_gstats(ctx, ETHTOOL_GSTATS, ETH_SS_STATS, "NIC");
}

static int do_gphystats(struct cmd_context *ctx)
{
	return do_gstats(ctx, ETHTOOL_GPHYSTATS, ETH_SS_PHY_STATS, "PHY");
}
```

stannous:
Yes, I remember 270c4018 where we fixed netsnmp_get_link_settings(). 
There, we used ETHTOOL_GLINKSETTIGS (if it existed) instead of ETHTOOL_GSET.
But even if we use ETHTOOL_GSTATS, we're still talking 10k ioctl calls (in my case).

agent/mibgroup/if-mib/data_access/interface.c
```diff
@@ -908,6 +940,31 @@ _free_interface_config(void)
...
+    name = strtok_r(cptr, " \t", &st);
+    if (!name) {
+        config_perror("Missing NUMBER parameter");
+        return;
+    }
+    long_val = strtol(cptr, NULL, 10);
+    if (errno) {
+           config_perror("Error converting parameter");
+           return;
+    }
```
bvanassche: From [https://pubs.opengroup.org/onlinepubs/009695399/functions/strtol.html](https://pubs.opengroup.org/onlinepubs/009695399/functions/strtol.html):

>     Upon successful completion, these functions shall return the converted value, if any. If no conversion
>     could be performed, 0 shall be returned [CX] [Option Start]  and errno may be set to [EINVAL].
> 
>     If the correct value is outside the range of representable values, {LONG_MIN}, {LONG_MAX}, {LLONG_MIN},
>     or {LLONG_MAX} shall be returned (according to the sign of the value), and errno set to [ERANGE].
> 

I think this means that it is not guaranteed that errno is set if strtol() fails. How about using sscanf() instead?

stannous: Yes, thanks.  I changed this to an sscanf in my update.

bvanassche:
The two patches have been combined into a single patch and have been applied on the v5.8 and master branches.

PR #49, opened 2019-12-04
Solve the memory leak in RHEL8

unequaled:
Based on the RHEL Bugzilla, Adding the rpmFreeRpmrc() before the exit function solves the memory leak if the function calls the rpmReadConfigFiles(). The sniff test at my side works fine.


hardaker:
Looks ok to me, thank you!

PR #56, opened 2020-01-07
Recognize NSS pools and NSS volumes (Open Enterprise Server)

tabraham:


bvanassche:

PR #63, opened 2020-02-04
libsnmp: Cast signed char to signed int when setting value

LeSpocky:
We have an INT8 in a MIB, an Integer32(-128..127) to be more specific.
In a custom software we call 'snmp_pdu_add_variable()' with
type == ASN_INTEGER and len == 1. Deeper down the code path, we come to
'snmp_set_var_value()' and this part:

        else if (vars->val_len == sizeof(char)) {
            if (ASN_INTEGER == vars->type) {
                const char      *val_char
                    = (const char *) value;
                *(vars->val.integer) = (long) *val_char;
            } else {
                    const u_char    *val_uchar
                    = (const u_char *) value;
                *(vars->val.integer) = (unsigned long) *val_uchar;
            }

Obviously the intent is to cast depending on whether the integer is
supposed to be signed or not, but although he jumps to the right code
path here, the result in vars->val.integer is wrong then. It does not
work with type 'char' like above.

We got values looking unsigned added to our pdu which we set to our
net-snmp server then and got them back unsigned on SNMP get requests
later, for example we fed an int8_t with value -32 we read back 224.

So this patch explicitly uses type 'signed char' which works for us.

Compiler is gcc 4.7 and we're cross-compiling for
arm-v5te-linux-gnueabi on Linux.

Signed-off-by: Alexander Dahl <ada@thorsis.com>

### DIFF
--- a/snmplib/snmp_client.c
+++ b/snmplib/snmp_client.c
@@ -888,8 +888,8 @@ snmp_set_var_value(netsnmp_variable_list * vars,
         }
         else if (vars->val_len == sizeof(char)) {
             if (ASN_INTEGER == vars->type) {
-                const char      *val_char 
-                    = (const char *) value;
+                const signed char   *val_char
+                    = (const signed char *) value;
                 *(vars->val.integer) = (long) *val_char;
             } else {
                     const u_char    *val_uchar
--- a/testing/fulltests/unit-tests/T023varbind_set_var_value_int8_clib.c
+++ b/testing/fulltests/unit-tests/T023varbind_set_var_value_int8_clib.c
@@ -1,0 +1,26 @@
+/* HEADER Testing snmp_set_var_value() */
+
+OK(1 == 1, "true");
+
+{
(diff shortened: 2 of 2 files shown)


LeSpocky:
I'm currently working on a unit test showing the initial failure, it's still in another branch of mine, though.

LeSpocky:
> I'm currently working on a unit test showing the initial failure, it's still in another branch of mine, though.

Unit test is ready and I rebased the fix on the other branch and force-pushed. You should see two commits in this PR now, the unit-test and the fix.

Note: I could not get the unit test to fail (before the fix) on x86 or x86_64. It fails on arm though, e.g. on Raspberry Pi.

LeSpocky:
- Added a Fixes: tag to the commit message.
- Confirm the unit-test is all green now on arm after the fix.

bvanassche:
Thanks for the patches! Both patches have been applied on the v5.8 and master branches.

Issue #64, opened 2020-02-06
errors during perl $sess->trap are not reported in $sess->{ErrorNum} and $sess->{ErrorStr}


pmorch:
First I show a reproduction recipe. Then a workaround. Finally a question asking for opinions on the best way forward to fix this.

## To reproduce

```
my $sess = new SNMP::TrapSession(
    DestHost => 'kosh',
    Community => 'public',
    Version => '2c',
);

# $SNMP::verbose = 1;

$sess->trap(
    oid => '.1.3.6.1.2.5',
    [
        ['bogus', 0, 1234],
    ]
);

printf "SNMP Error? ErrorNum: %d: ErrorStr: %s\n",
    $sess->{ErrorNum},
    $sess->{ErrorStr} eq '' ? "<empty string - no error>" : $sess->{ErrorStr};
```

This prints output showing that using the standard API it is not possible to detect that issuing the trap has failed (as it should since `bogus` is not a valid OID):

    SNMP Error? ErrorNum: 0: ErrorStr: <empty string - no error>

## Workaround

If `$SNMP::verbose`, then a warning is generated. Hopefully in every case :-) . So this workaround is possible:

```
{
    local $SNMP::verbose = 1;
    my $warning;
    local $SIG{__WARN__} = sub {
        my ($w) = @_;
        $warning = $w;
    };
    $sess->trap(
        oid => '.1.3.6.1.2.5',
        [
            ['bogus', 0, 1234],
        ]
    );
    if ($warning) {
        # do something with $warning as if it was in $sess->{ErrorStr}
        print "Got warning during \$sess->trap: " . $warning;
    }
}
```

## What is the best way to fix this?

Should one introduce perl-specific error codes in `snmp_api.h` and `snmp_api.c`'s `static const char *api_errors[-SNMPERR_MAX + 1]` for `$sess->{ErrorNum}` each individual error case, or should one guess the most relevant error code, falling back to SNMPERR_GENERR and then put the string that is already used for the warning in `$sess->{ErrorStr}`?

The standard way to handle errors in the perl API is to set `$sess->{ErrorNum}` and `$sess->{ErrorStr}`. In `SNMP.xs` this usually looks like this:

    sv_setpv(*err_str_svp, (char*)snmp_api_errstring(SNMPERR_TIMEOUT));
    sv_setiv(*err_num_svp, SNMPERR_TIMEOUT);

[...]

PR #65, opened 2020-02-06
Perl can notify empty octet string

pmorch:
This works:

```
my $sess = new SNMP::TrapSession(
    DestHost => 'kosh',
    Community => 'public',
    Version => '2c',
);

# $SNMP::verbose = 1;

$sess->trap(
    oid => '.1.3.6.1.2.5',
    [
        ['sysLocation', 0, "here"],
    ]
);

```

If I change the `sysLocation` to the empty string, no notification is sent, but because of #64, `$sess->{ErrorNum}` and `$sess->{ErrorStr}` don't indicate any problem. Setting `$SNMP::verbose = 1;` yields this warning:

     error:trap v2: adding varbind at /usr/lib/x86_64-linux-gnu/perl5/5.24/SNMP.pm line 1292.

This PR makes it possible to send traps with OCTET STRING-s of length 0, just like you get with:

    snmptrap -v2c -cpublic kosh '' 1.3.6.1.2.5 sysLocation.0 s ''

### DIFF
--- a/perl/SNMP/SNMP.xs
+++ b/perl/SNMP/SNMP.xs
@@ -1022,11 +1022,11 @@ as_uint:
       case TYPE_OPAQUE:
         vars->type = ASN_OCTET_STR;
 as_oct:
-        vars->val.string = netsnmp_malloc(len);
-        vars->val_len = len;
-        if (val && len)
+        if (val && len) {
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for these patches! Modified versions of both patches have been applied on the v5.8 and master branches. Please retest.

PR #70, opened 2020-02-20
Update README

harsha-s:
Added contributor for contributions in https://sourceforge.net/projects/net-snmp/

https://sourceforge.net/p/net-snmp/patches/1224/


hardaker:
So I've actually removed the list entirely from the README file as it's horribly out of date.  instead I put in text pointing to the changelog file and the git history, which contains the name in the proposed patch.  I wish there was an easier way to track contributions to long term projects.  Git makes it easier, since pull requests directly incorporated should now appear as authored by someone as opposed to our older way of having to manually put in information into log messages.

PR #73, opened 2020-02-21
avoid snmpps null deref for SNMP_NOSUCHOBJECT

sthen:
Handle SNMP_NOSUCHOBJECT as well as SNMP_NOSUCHINSTANCE. If snmpps/snmptop are pointed at an agent which doesn't provide the required objects, before this change it tries to dereference a null pointer when setting the variables (noticed against OpenBSD snmpd which only implements part of hrSWRunPerf and misses CPU/Mem).


bvanassche:
Thanks for the patch. This patch has been applied on the v5.8 and master branches.

PR #74, opened 2020-02-21
OpenBSD inpt_queue uses TAILQ not CIRCLEQ

sthen:
OpenBSD changed from CIRCLEQ to TAILQ for inpt_queue in OpenBSD 5.6 (November 2014), the version of Net-SNMP in ports has been carrying this patch since then.


bvanassche:
Thanks for the patch. This patch has been applied on the v5.8 and master branches.

Issue #58, opened 2020-01-17
Memory leak in handle_agentx_packet when using a contextName


jain-mukesh:
Our infrastructure code uses net-snmp and I found a memory leak in handle_agentx_packet if contextName is given.

Below is the trace from valgrind

```==30360== 15 bytes in 5 blocks are definitely lost in loss record 492 of 1,919
==30360==    at 0x4C29EA3: malloc (vg_replace_malloc.c:309)
==30360==    by 0x653F294: snmp_clone_mem (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x653F3E7: ??? (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x653F90D: snmp_clone_pdu (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x60E3837: handle_agentx_packet (in /usr/lib64/libnetsnmpagent.so.31.0.2)
==30360==    by 0x6567ED0: ??? (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x6568F40: _sess_read (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x65698A8: snmp_sess_read2 (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x65698FA: snmp_read2 (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x656993B: snmp_read (in /usr/lib64/libnetsnmp.so.31.0.2)
==30360==    by 0x60DF30A: agent_check_and_process (in /usr/lib64/libnetsnmpagent.so.31.0.2)
```

On further debugging, we found that the internalPDU->contextName was assigned to the internalPDU->community resulting in losing the reference to the assigned memory of internalPDU->contextName. This memory is never free'ed when the internalPDU is deallocated. 

Proposed fix:
contextNames memory should be freed before assignment to community, so that the memory is not lost.

In agent/mibgroup/agentx/subagent.c

`   internal_pdu = snmp_clone_pdu(pdu);`

> `SNMP_FREE(internal_pdu->contextName);`

`    internal_pdu->contextName = (char *) internal_pdu->community;`
`    internal_pdu->contextNameLen = internal_pdu->community_len;`

bvanassche:
[...]

PR #78, opened 2020-03-02
Handle a memory leak when contextName is used in the PDU. The memory …

jain-mukesh:
We found that the internalPDU->contextName was assigned to the internalPDU->community resulting in losing the reference to the assigned memory of internalPDU->contextName. This memory is never free'ed when the internalPDU is deallocated. I tested the issue in 5.3, 5.8 and master branch. The issue does not happen if contextName is not used

Proposed fix:
contextNames memory should be freed before assignment to community, so that the memory is not lost.

This is discussed further in issue #58 along with a valgrind report.


bvanassche:
A slightly modified patch has been applied on the v5.8 and master branches. Thanks for the patch!

PR #83, opened 2020-03-18
mib2c: use struct counter64 instead of U64 in generated code

idryzhov:
The U64 typedef was replaced with struct counter64 in net-snmp
code because of the conflicts with Perl headers. But mib2c still
generates the code with U64.

Signed-off-by: Igor Ryzhov <iryzhov@nfware.com>


bvanassche:
This patch has been applied on the v5.8 and master branches. Thanks!

PR #84, opened 2020-03-20
mib2c: fix size_t format specifier in mfd generated code

idryzhov:
Signed-off-by: Igor Ryzhov <iryzhov@nfware.com>

### DIFF
--- a/local/mib2c-conf.d/mfd-access-container-cached-defines.m2i
+++ b/local/mib2c-conf.d/mfd-access-container-cached-defines.m2i
@@ -358,7 +358,7 @@ $example_end
 @end@ # example
 
     DEBUGMSGT(("verbose:${context}:${context}_container_load",
-               "inserted %d records\n", count));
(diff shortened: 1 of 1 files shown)


local/mib2c-conf.d/mfd-access-container-cached-defines.m2i
```diff
@@ -358,7 +358,7 @@ $example_end
 @end@ # example
 
     DEBUGMSGT(("verbose:${context}:${context}_container_load",
-               "inserted %d records\n", count));
+               "inserted %zu records\n", count));
```
bvanassche: How about using NETSNMP_PRIz instead of 'z'? The 'z' format specifier is supported by glibc but not by all versions of Microsoft Visual Studio.

idryzhov: Done.

bvanassche:

PR #85, opened 2020-03-21
Add regex support for include_ifmib_iface_prefix

lbjw:
This patch builds upon commit 76ad71101ce0a4db50f5ab9dd025b13874456b55 which added the include_ifmib_iface_prefix configuration option to instead utilise regular expressions rather than string comparisons. The use case is a common situation where you might have thousands of interfaces named like eth3.stag.ctag and you still want to SNMP monitor eth3 but not all of the other interfaces. Specifying eth3 in the current code would cause all the interfaces to be picked up.


bvanassche:
A slightly modified patch has been applied on the v5.8 and master branches. Thanks for the patches!

PR #91, opened 2020-04-04
swinst_apt: Use dpkg for modified date

csmall:
Instead of trawling through dpkg's files, let dpkg-query tell us
when the package was last modified. It means if the directory
changes or any other internal changes happen, this will still
keep working.

db-sys:Last-Modified field has been in since dpkg-query 1.19.3
which means it works for Debian stable (Buster) onwards.

References:
 https://bugs.debian.org/905668

### DIFF
--- a/agent/mibgroup/host/data_access/swinst_apt.c
+++ b/agent/mibgroup/host/data_access/swinst_apt.c
@@ -33,17 +33,14 @@
 
 config_require(date_n_time)
 
-char pkg_directory[SNMP_MAXBUF];
 static char apt_fmt[SNMP_MAXBUF];
-static char file[SNMP_MAXBUF];
 
 /* ---------------------------------------------------------------------
  */
 void
 netsnmp_swinst_arch_init(void)
 {
-    strlcpy(pkg_directory, "/var/lib/dpkg/info", sizeof(pkg_directory));
-    snprintf(apt_fmt, SNMP_MAXBUF, "%%%d[^#]#%%%d[^#]#%%%d[^#]#%%%d[^#]#%%%d[^#]#%%%d[^#]#%%%ds",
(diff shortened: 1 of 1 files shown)


csmall:
What does it look like? It looks the same! Stock Debian snmpd is listening on port 161, patched snmpd is listening on port 9161
```
$ snmpget -v 1 -c public localhost:161 1.3.6.1.2.1.25.6.3.1.5.505 1.3.6.1.2.1.25.6.3.1.2.505
HOST-RESOURCES-MIB::hrSWInstalledDate.505 = STRING: 2020-3-18,20:58:24.0,+11:0
HOST-RESOURCES-MIB::hrSWInstalledName.505 = STRING: "libc6_2.30-2_amd64"

$ snmpget -v 1 -c public localhost:9161 1.3.6.1.2.1.25.6.3.1.5.505 1.3.6.1.2.1.25.6.3.1.2.505
HOST-RESOURCES-MIB::hrSWInstalledDate.505 = STRING: 2020-3-18,20:58:24.0,+11:0
HOST-RESOURCES-MIB::hrSWInstalledName.505 = STRING: "libc6_2.30-2_amd64"
```

bvanassche:
A slightly modified patch has been checked in on the v5.8 and master branches. Please retest.

csmall:
As the modified_date and modified_date_t are not pointers, won't this mean the value is not updated?  The assignment would need to happen after the scanf but before you set assign it to the new entry.

An alternative would be to make modified_date_t a pointer to modified_date as they both will update at the same time.

bvanassche:
Right, I should have been more careful when I made my change. I will revert the patch that went in.

There is another issue with this patch: if I run the command from the patch in a shell, no last-modified date is displayed:
~~~
$ cat /etc/issue.net 
Debian GNU/Linux 9
$ dpkg-query --show --showformat \
'${Package}#${Version}#${Section}#${Priority}#${Essential}#${Architecture}#${Status}#${db-fsys:Last-Modified}\n' libc-bin
libc-bin#2.24-11+deb9u4#libs#required#yes#amd64#install ok installed#
~~~

PR #121, opened 2020-05-27
V5 7 patches

lmbzx:
On HPUX 11.31  snmpwalk over mibII get timeout if the number of interfaces of the server is large.
This is due to the excessive number of get_phys_stat() called, causing the stats to be get thousand of time for a simple walk

With this fix, get_phys_stat is only get when  SaveIndex is 0 so, starting a new lookup on an interface.


bvanassche:
The v5.7 branch is no longer maintained. The active branches are the v5.8 and the master branch.

Additionally, the mibII/interfaces.[ch] implementation is obsolete. Please use if-mib/ifTable instead. It shouldn't be that hard to add a agent/mibgroup/if-mib/data_access/interface_hpux.c implementation.

PR #122, opened 2020-05-27
fix excessive call to get_phys_stat on hpux causing huge delay when n…

lmbzx:
On HPUX 11.31 snmpwalk over mibII get timeout if the number of interfaces of the server is large.
This is due to the excessive number of get_phys_stat() called, causing the stats to be get thousand of time for a simple walk

With this fix, get_phys_stat is only get when SaveIndex is 0 so, starting a new lookup on an interface.


hardaker:
I'm not sure this patch will work when two snmpwalks are running against the table at the same time.  I think the right thing to do is introduce time-based data caching within the agent, like other mib implementations use.

Issue #128, opened 2020-06-01
Unnecessary use of keyword register prevent compiling with newer C++ versions


jbulow:
The use of the keyword "register" causes error when compiling with newer C++ standards (C++17 and higher). 

Also, the use of the register keyword does not add any value.

https://github.com/net-snmp/net-snmp/blob/41553b0cb6cf4ccca7edf7eca471b46724664e6c/include/net-snmp/library/mib.h#L484

hardaker:
Yeah, it's long past time.  "back in my day" it actually did make a difference.

PR #129, opened 2020-06-01
Remove obsolete keyword.

jbulow:
This will allow compilation using C++17 where the register keyword is obsoleted.

Also, the register keyword is in general never necessary.


jbulow:
Fixes #128

bvanassche:
Thanks for the patch!

PR #146, opened 2020-07-09
sensors: Add support for humidity sensor

chaitu-tk:
LM-SENSORS-MIB supports only 4 standard sensors but provides MISC table
for other sensors, we can use that table to log the humidity sensor
reading.

Tested using my local unit and works fine.


chaitu-tk:
We could as well add others to the Misc table, but I was only concerned with humidity.

hardaker:
Thanks for the patch!

Issue #147, opened 2020-07-09
Net-SNMP not responding when proxy requests times out


fisabelle:
Hi,
 
Our Net-SNMP daemon is configured to proxy requests on another agent. If this agent is not running, or in case of internal failure or timeout in the sub agent, the proxy requests fail as expected. But if a snmpset query is executed afterwards, the set request is queued and never processed. This causes all other subsequent snmpget requests to be never processed as well. That leaves our Net-SNMP daemon not responsive for ever.
 
Setup:
======
To easily reproduce, a fake agent can be used in the proxy directive.
Our Net-SNMP daemon is running on port 8001. Net-SNMP is at latest revision (rev5a46404) from V5-8-patches stream
Proxy directive in our snmpd.conf: proxy -v 2c -c public localhost:163 .1.3.6.1.6666 (note here that in this test, there is no agent running on port 163)
 
How to reproduce
======
Execute a snmpget on our Net-SNMP that will be sent to the target agent.
```
snmpget -v 2c -c public 127.0.0.1:8001 .1.3.6.1.6666.1.0
``` 
This request times out as we never get a response.
 
Execute a snmpget.
```
snmpget -v 2c -c public 127.0.0.1:8001 .1.3.6.1.2.1.11.30.0
```
This request is successful as we get the expected results.
 
Execute a snmpset.
```
snmpset -v 2c -c public 127.0.0.1:8001 .1.3.6.1.2.1.11.30.0 i 1
```
This request times out and is never processed.
 
Execute a snmpget
```
snmpget -v 2c -c public 127.0.0.1:8001 .1.3.6.1.2.1.11.30.0
```
This request times out. At this moment, Net-SNMP is not responding anymore to any requests.
 
From the log file (attached), it seems that the set request is queued because it thinks the snmpget that was proxied is not finished. All subsequent snmpget requests are queued because there is a set request queued.
 
This looks like a bug where the proxy request is not cleared from the delegate list. Is it possible to have a fix for this?
[snmpd2020.07.09-09.50.35.log.gz](https://github.com/net-snmp/net-snmp/files/4897334/snmpd2020.07.09-09.50.35.log.gz)

hardaker:
Fixed by PR #147 (hopefully).

PR #148, opened 2020-07-09
Issue#147: Net-SNMP not responding when proxy requests times out

fisabelle:
This PR resolves the issue #147.

The test scenario presented in the issue no longer fails with this fix in place.
Without a complete understanding of the actual expected flow, I can not tell if there are any side effects such as leaks or locks in other scenarios.

A short explanation of the fix intent is provided in the code.


hardaker:
Thanks for the patch; it does look like a potentially missing case element.

PR #163, opened 2020-08-17
snmpd: set NULL to rp->cb_data for avoid access already freed area

daisuke-kaguchi:


bvanassche:
Please reopen after a reproducer has been provided.

PR #164, opened 2020-08-17
snmpd: set NULL to rp->cb_data for avoid access already freed area

daisuke-kaguchi:


daisuke-kaguchi:
The symptom occurred when the snmpwalk command was repeatedly executed for about two days.

bvanassche:
Please reopen after a reproducer has been provided.

PR #167, opened 2020-08-21
snmpd: support MemAvailable on Linux

ibigbug:
This is my first time walking into this code base, so I don't know what I'm doing.

But the idea is to add support of reporting MemAvailable on Linux

```
-> % free -h
              total        used        free      shared  buff/cache   **available**
Mem:          3.8Gi       437Mi       2.5Gi       5.0Mi       936Mi       3.2Gi
Swap:         4.0Gi          0B       4.0Gi
```

please guide if I'm doing this correctly and/or anything that is not covered

### DIFF
--- a/agent/mibgroup/hardware/memory/memory_linux.c
+++ b/agent/mibgroup/hardware/memory/memory_linux.c
@@ -24,7 +24,7 @@ int netsnmp_mem_arch_load( netsnmp_cache *cache, void *magic ) {
     static int   first = 1;
     ssize_t      bytes_read;
     char        *b;
-    unsigned long memtotal = 0,  memfree = 0, memshared = 0,
+    unsigned long memtotal = 0,  memavail = 0, memfree = 0, memshared = 0,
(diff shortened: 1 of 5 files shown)


csmall:
I'm not sure what the RFC says but putting my procps (free, top) hat on for a second the available figure is the most useful. There is a commit floating around somewhere from Linus about why this is a good value to use for memory pressure (e.g. how much more memory can I allocate before causing memory issues such as swapping).

ibigbug:
done ping @bvanassche

bvanassche:
A modified version of this patch has been applied. Please take a look at commit  5b8bf5d4130761c3374f9ad618e8a76bb75eb634 .

ibigbug:
thanks @bvanassche 

may I know when will this patch be released?

bvanassche:
The date of the next release has not yet been decided as far as I know.

ibigbug:
@bvanassche any chance to have a release recently? so this can be leveraged by https://github.com/librenms/librenms/pull/12110

PR #174, opened 2020-08-27
snmpd: avoid container double free in error cases

mark123499:
container should be freed by caller function netsnmp_access_interface_container_load().


bvanassche:
This patch has been applied on the v5.9 and master branches. Thanks for the patch!

Issue #135, opened 2020-06-13
Perl memory leak with Net::SNMP gettable() method


fellowsr:
# Tested on - 
 * net-snmp 5.5 on Oracle Linux 6, vendor package
 * net-snmp 5.7.3 on Raspian 4.19.118-v7+, vendor package
 * net-snmp 5.8 on Ubuntu 20.04 LTS, build from source.
# Replication steps
Run the perl script below and monitor memory usage with `top` or `htop`.

```perl
#!/usr/bin/perl
use strict;
use warnings;
use SNMP;
my $session = SNMP::Session->new(
     DestHost  => '127.0.0.1',
     Version   => 2,
     Community => 'secret'
);
for ( 0 .. 20000 ) {
     my $if = $session->gettable('ifTable');
}
__END__
```
htop output a few seconds apart -
```
   PID USER      PRI  NI  VIRT   RES   SHR S CPU% MEM%   TIME+  Command
  83596 fellowsr   20   0 21504 16776  6548 R 84.1  0.8  0:06.71 perl ./gettable.pl
```
later -
```
    PID USER      PRI  NI  VIRT   RES   SHR S CPU% MEM%   TIME+  Command
  83596 fellowsr   20   0 46608 41880  6548 R 85.2  2.1  0:43.28 perl ./gettable.pl
```
I've written similar test scripts for other methods defined in SNMP.pm https://metacpan.org/pod/SNMP

* get() - no leak.
* fget() - no leak.
* fgetnext() - no leak.
* getnext() - no leak.
* gettable with various calls -
  * `$session->gettable('ifTable')` - **leaks**.
  * `$session->gettable('ifTable')` - **leaks.**
  * `$session->gettable('ifTable',callback => \&my_cb);` **leaks**
  * `$session->gettable('ifTable', column => [ 'ifInOctets' ]);` **leaks**
  * `$session->gettable('ifTable',nogetbulk =>1);` **leaks**
  * `$session->gettable( 'ifTable', noindexes => 1 );` **leaks**
* trap - no leak.

# Edit 1
Happy to run valgrind or other tools to attempt to tie this down. Just tell me what you need me to run. I've been trying to do this myself but I've not done any prior work with XS code, so the results don't make a lot of sense to me.

# Edit 2
Had more success with valgrind I think?

```bash
[...]

PR #179, opened 2020-09-02
fix resource leaks in the SNMP::Session destructor and SNMP::NODE::FETCH

PavelTrushkin:
# leak in SNMP::Session destructor

```c
void
snmp_session_DESTROY(sess_ptr)
        void *sess_ptr
```

in this block sess_ptr MUST be defined as SnmpSession *

```c
void
snmp_session_DESTROY(sess_ptr)
        SnmpSession *sess_ptr
```

here is a results of conversion .xs to .c for this code

for void *

```c
void *  sess_ptr = INT2PTR(void *,SvIV(ST(0)))
```

for SnmpSession *

```c
       SnmpSession *   sess_ptr;
   
        if (SvROK(ST(0))) {
            IV tmp = SvIV((SV*)SvRV(ST(0)));
            sess_ptr = INT2PTR(SnmpSession *,tmp);
        }
        else
            Perl_croak_nocontext("%s: %s is not a reference",
                        "SnmpSessionPtr::DESTROY",
                        "sess_ptr")
```

destructor calls snmp_close or snmp_sess_close with absolutely wrong parameter and we have a huge resource leak

# two leaks in snmp_mib_node_FETCH

code creates new SV and RV but uses it only inside the comparison branch

# how to reproduce

this problem can be reproduced using script and command line from #135

### DIFF
--- a/perl/SNMP/SNMP.xs
+++ b/perl/SNMP/SNMP.xs
@@ -5190,11 +5190,11 @@ snmp_mib_node_FETCH(tp_ref, key)
                     mib_hv = perl_get_hv("SNMP::MIB", FALSE);
                     if (SvMAGICAL(mib_hv)) mg = mg_find((SV*)mib_hv, 'P');
                     if (mg) mib_tied_href = (SV*)mg->mg_obj;
-                    next_node_href = newRV((SV*)newHV());
                     __tp_sprint_num_objid(str_buf, tp);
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patches!

PavelTrushkin:
please read my comment to commit bad073bd860e5a4ae805747b7ce043aea784d9de

it is very important !

PR #180, opened 2020-09-03
Parse MIBs: Raise limit for external imports

gjaekel:
Today, I stepped into an well-known old issue (https://sourceforge.net/p/net-snmp/mailman/message/16927159/) using a bunch of MIBs from CISCO.

### DIFF
--- a/snmplib/parse.c
+++ b/snmplib/parse.c
@@ -28,7 +28,7 @@ ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS
 SOFTWARE.
 ******************************************************************/
 /*
- * Copyright © 2003 Sun Microsystems, Inc. All rights reserved.
+ * Copyright Â© 2003 Sun Microsystems, Inc. All rights reserved.
  * Use is subject to license terms specified in the COPYING file
  * distributed with the Net-SNMP package.
  */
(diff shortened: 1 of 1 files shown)


bvanassche:
A slightly modified version of this patch has been applied as commit https://github.com/net-snmp/net-snmp/commit/893b9baa6ae81a877e76842c29a637eeb2cbd0e2. Please take a look.

gjaekel:
Hi Bart,

that's all right, GitHub.com seems to made an autoconvert of the code page to UTF-8. Because of the "local" position of the `#define` it seems to me that in former times somebody have done a small refactoring here.

Thank you for accepting the patch immediately
Guido

bvanassche:
Since allocating large arrays on the stack is not optimal, how about reworking the MIB parser such that that array is allocated dynamically instead of on the stack?

gjaekel:
Because of the pressure to get the usecase running, I don't take a closer look at the code to get notice of this. In spite of having a background of more than 35 years and can read and understand code written in many languages, I'm currently not act as an developer. Therefore I neither have a ready toolchain nor much deeper handy experience in writing code. A plumber like you would write it down in less than a half hour.

I'm using Gentoo and with this I'm able to figure out such simple things just by using vi and other standard Unix tools. I love the way GitHub offers for for people like me to propose small changes just using a browser. And I know what kind of changes might be proposed even without a full build. But a refactoring like the one you propose should be made with the support of an IDE and confirmed by testcases, too. Therefore, I have to leave this as a ToDo, sorry for that.

PR #183, opened 2020-09-16
Fix memory leak in ipAddressPrefixTable_data_access.c

urs-k:
The memory was leaked when, while populating ipAddressPrefixTable, "do we already have this prefix?" was true and no CONTAINER_INSERT happened afterwards. On my test system this always happened. In addition, the iterator was not released if memory allocation failed later, and the result of CONTAINER_INSERT was not checked.


bvanassche:
A slightly modified patch has been applied on the v5.9 and master branches. Thanks for the patch!

PR #197, opened 2020-10-07
Fix regex for "-l" discovery in setup.py

haxtibal:
This prevents setup.py from discovering wrong library link options (`-l`, `-L`) in the middle of a word, like `"inux-gnu"` from `"-L/usr/lib/x86_64-linux-gnu"`.

For `net-snmp-config` output as follows (real world example from Ubuntu 18.04):
```
-Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now -L/usr/lib/x86_64-linux-gnu -lnetsnmp -lcrypto -lm
```
the former regex `-l(\S+)` would match to
```python
['inux-gnu', 'netsnmp', 'crypto', 'm']
```
which obviously results in a ld error `cannot find -linux-gnu`.

Now we capture a string only, if `-l` is at the beginning of a string, or if `-l` is preceded by any number of whitespace characters.


hardaker:
Looks good; thanks.

PR #203, opened 2020-11-09
agent: expose netsnmp_agent_request_info_s before snmp feeding

pguibert6WIND:
The snmp api v4 API to register mib does not provide any parameter
to know which session the mib is for. One way to mitigate this
limitation is to create a callback service, so that the
netsnmp_agent_request_info_s is available. This permits to know the used
netsnmp_agent_session_s, and the underlying netsnmp_session used.

This solution is not perfect, but may be a correct trade off, assuming that
the transition from old v4 API code to new v5 API may be not neglectible.

Signed-off-by: Philippe Guibert <philippe.guibert@6wind.com>


pguibert6WIND:
any updates on this point, please?

hardaker:
Looks generally ok.  I wonder about its use in multi-threaded applications, so be wary as I haven't studied that.

PR #204, opened 2020-11-13
mib2c-conf: add support for in_addr_t

pguibert6WIND:
when a mib table contains ASN_IPADDRESS references, the mib2c compiler
fails to produce adequate skeleton files, and indicates this message:

ERROR: unknown node.decl: in_addr_t

Fix this by adding support for in_addr_t.

Signed-off-by: Philippe Guibert <philippe.guibert@6wind.com>


local/mib2c-conf.d/generic-get-in_addr_t.m2i
```diff
@@ -0,0 +1,13 @@
+#############################################################  -*- c -*-
+## generic include for XXX. Do not use directly.
```
bvanassche, line 2: Should "XXX" perhaps be changed into "in_addr_t"?

pguibert6WIND: sure..

bvanassche:
Thanks for the patch!

PR #206, opened 2020-11-13
Include inactive memory with cache under FreeBSD; Fixes #205

### LINKED ISSUE #205: Memory reported incorrectly under FreeBSD
There is a longstanding issue where net-snmp makes it impossible to correctly instrument "used" and "free" memory in FreeBSD due to limitations of the UCD-SNMP-MIB. For many years the FreeBSD port has patched in a fix for this by reporting "inactive" memory along with "cached" in the `UCD-SNMP-MIB::memCached` MIB. This makes it possible to report on "used" memory via (memTotalReal - memAvailReal - memCached).

Unfortunately this patch was never upstreamed to net-snmp and was recently removed from the port.

I have opened an issue with the FreeBSD ports team which describes the problem in more details:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=251103

Would it be possible to get this upstreamed?

courtland:
This fixes a memory reporting problem under FreeBSD, which the FreeBSD port has relied on as a patch for many years.

### DIFF
--- a/agent/mibgroup/hardware/memory/memory_freebsd.c
+++ b/agent/mibgroup/hardware/memory/memory_freebsd.c
@@ -164,7 +164,7 @@ int netsnmp_mem_arch_load( netsnmp_cache *cache, void *magic ) {
         if (!mem->descr)
              mem->descr = strdup("Cached memory");
         mem->units = pagesize;
-        mem->size  = cache_count;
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

Issue #220, opened 2020-12-04
EngineBoots increments after config reload


dmoricon:
Hi,
I have noticed that everytime I reload the configuration, the engineBoots increment. I am not sure if this is the correct behavior. Plus, the engineTime is not reset when the engineBoots is incremented as per documented in the RFC. 

`
snmpget -v 2c -c public 127.0.0.1:161 .1.3.6.1.6.3.10.2.1.2.0
SNMP-FRAMEWORK-MIB::snmpEngineBoots.0 = INTEGER: 637
snmpget -v 2c -c public 127.0.0.1:161.1.3.6.1.6.3.10.2.1.3.0
SNMP-FRAMEWORK-MIB::snmpEngineTime.0 = INTEGER: 107 seconds
snmpset -v 2c -c public 127.0.0.1:161 .1.3.6.1.4.1.2021.100.11.0 i 1
UCD-SNMP-MIB::versionUpdateConfig.0 = INTEGER: 1
snmpget -v 2c -c public 127.0.0.1:161.1.3.6.1.6.3.10.2.1.2.0
SNMP-FRAMEWORK-MIB::snmpEngineBoots.0 = INTEGER: 638
snmpget -v 2c -c public 127.0.0.1:.1.3.6.1.6.3.10.2.1.3.0
SNMP-FRAMEWORK-MIB::snmpEngineTime.0 = INTEGER: 122 seconds
`

hardaker:
copying response to the PR here too:

So we actually did this intentionally, though the exact case that cause us to do this escapes my memory. But in essence a re-read of the config could change many aspects of the agent and is functionally a "reboot" in that sense. Thus, it is safer to increment the boots value as the only real supposed downside in v3 is that there will be an extra round trip for managers to update their notions of it. And re-reads of the configuration should be rare.

hardaker:
That being said, I do agree that the engineTime should be reset.

iyyapa:
I was testing the Snmpv3 traps and found that the USM time synchronization is failing.
Reload the switch, upon booting the SLX will send the traps to the ubuntu server and from the packet capture we can see the msgAuthoritativeEngineBoots is always 1.

According to RFC 3414 the value should be persist even after reboot. It should increment every reboot.
Kindly let me know, if this got fixed in any of the patches.

fenner:
[...]

PR #222, opened 2020-12-04
Do not increment EngineBoots on config reload

dmoricon:
This is a pull request for Issue #220 to avoid incrementing the EngineBoots when the config is reloaded. The following use cases were tested
1- Restart: Make sure the engineBoots is incremented.
2- Config reload: Make sure the engineBoots is not incremented.

This is for V5-8-patches since we are still using this version. Maybe it is also needed in 5-9 as well.

### DIFF
--- a/snmplib/snmpv3.c
+++ b/snmplib/snmpv3.c
@@ -994,7 +994,7 @@ init_snmpv3(const char *type)
                                     engineIDType_conf, NULL, "num");
     register_prenetsnmp_mib_handler(type, "engineIDNic", engineIDNic_conf,
                                     NULL, "string");
-    register_config_handler(type, "engineBoots", engineBoots_conf, NULL,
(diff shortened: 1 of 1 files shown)


hardaker:
So we actually did this intentionally, though the exact case that cause us to do this escapes my memory.  But in essence a re-read of the config could change many aspects of the agent and is functionally a "reboot" in that sense.  Thus, it is safer to increment the boots value as the only real supposed downside in v3 is that there will be an extra round trip for managers to update their notions of it.  And re-reads of the configuration should be rare.

fenner:
I know I'm late to this party, but just wanted to mention that if you're logging to a file with `-Lf`, in order to rotate that log file you have to HUP snmpd, so it happens just as often as you need to rotate the log file, even if the config hasn't changed at all.

PR #224, opened 2020-12-08
perl binding: add strong authentication and privacy

garnier-quentin:
We can use strong authentication (SHA256, SHA512,...) with perl bindings. I also done some indent change (it was the mess in the method).


bvanassche:
A modified version of these patches has been applied on the v5.9 and master branches. Please take a look.

garnier-quentin:
Thanks. Could it be merged in 5.8 branch also ?

bvanassche:
The team that maintains Net-SNMP is small so we only support the v5.9 and master branches.

PR #227, opened 2020-12-13
Fix warnings generated by configure scripts

mitchblank:
Specifically, this is meant to reduce risk of issues with Xcode 12 which has made `-Werror,-Wimplicit-function-declaration` a default.

Normally this doesn't bite "real" code since such a warning would be visible for a long time with most compilers.  However, inside of configure scripts these sorts of warnings just end up in config.log and usually get ignored.  Now that they produce errors it means that configure tests that *should* pass suddenly fail, causing very strange failures later on.

This can be as simple as calling `exit(0);` inside of a test C program without doing `#include <stdlib.h>` first!

I also fixed an implicit-int-return warning since that is another C behavior that was deprecated long ago and future compilers might consider that an error.

### DIFF
--- a/configure.d/config_os_functions
+++ b/configure.d/config_os_functions
@@ -212,12 +212,12 @@ AC_CACHE_VAL(
 #ifdef HAVE_SYS_FS_TYPES_H
 #include <sys/fs_types.h>
 #endif
-main ()
+int main ()
 {
 struct fs_data fsd;
 /* Ultrix's statfs returns 1 for success,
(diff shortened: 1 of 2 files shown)


bvanassche:
Although I concerned that enabling `-Werror,-Wimplicit-function-declaration` by default will break all configure scripts, including the Net-SNMP configure script with this patch applied, I have applied this patch.

bvanassche:
See also commit a7c8c26c48c954a19bca5fdc6ba285396610d7aa.

PR #235, opened 2020-12-17
Backport: Increase extension buffer size, avoid a crash

minfrin:
Make sure enough space is allocated for extensions longer than 512 bytes. Fix
the error message when the extension is too long to clearly differentiate the
error from the out of memory case. Fix a crash when the extension is too long.


minfrin:
Grrr... wrong branch.

Issue #187, opened 2020-09-24
snmpd: get_target_sessions returns already deleted session after killing snmptrap 


emilbart:
I found one problem when running snmpd (from net-snmp version is 5.8 and on ubuntu) under valgrind:

```
==22001== Invalid read of size 8
==22001==    at 0x5112202: get_target_sessions (target.c:302)
==22001==    by 0x5140A77: send_notifications (snmpNotifyTable_data.c:222)
==22001==    by 0x557E7F5: snmp_call_callbacks (callback.c:363)
==22001==    by 0x4E760B5: netsnmp_send_traps (agent_trap.c:1033)
==22001==    by 0x4E76128: send_enterprise_trap_vars (agent_trap.c:1047)
==22001==    by 0x4E77223: send_trap_vars (agent_trap.c:1311)
==22001==    by 0x4E99BB4: agentx_notify (master_admin.c:456)
==22001==    by 0x4E9A116: handle_master_agentx_packet (master_admin.c:554)
==22001==    by 0x5551C5B: _sess_process_packet_handle_pdu (snmp_api.c:5804)
==22001==    by 0x5551DE2: _sess_process_packet (snmp_api.c:5860)
==22001==    by 0x55533B3: _sess_read (snmp_api.c:6286)
==22001==    by 0x555388D: snmp_sess_read2 (snmp_api.c:6394)
==22001==  Address 0x7a8b9b0 is 448 bytes inside a block of size 488 free'd
==22001==    at 0x4C30D3B: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==22001==    by 0x5543450: snmp_free_session (snmp_api.c:1931)
==22001==    by 0x5543723: snmp_sess_close (snmp_api.c:2016)
==22001==    by 0x5543802: snmp_close (snmp_api.c:2046)
==22001==    by 0x5553CE7: snmp_sess_select_info2_flags (snmp_api.c:6556)
==22001==    by 0x5553ACF: snmp_sess_select_info2 (snmp_api.c:6486)
==22001==    by 0x555393C: snmp_select_info2 (snmp_api.c:6442)
==22001==    by 0x10C7CD: receive (snmpd.c:1262)
==22001==    by 0x10C372: main (snmpd.c:1125)
==22001==  Block was alloc'd at
==22001==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

PR #237, opened 2020-12-17
Fix the crash of snmpd when snmptrapd terminates the TCP connection.

emilbart:
See also: https://github.com/net-snmp/net-snmp/issues/187

Re-opening of TCP connection with snmptrapd is still not supported.

### DIFF
--- a/agent/agent_trap.c
+++ b/agent/agent_trap.c
@@ -183,9 +183,9 @@ _dump_trap_stats(netsnmp_session *sess)
 #endif /* NETSNMP_NO_TRAP_STATS */
 
 int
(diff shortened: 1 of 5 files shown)


emilbart:
This patch is mainly intended as a rough draft.

hardaker:
unfortunately, this changes a published API and thus can't be accepted as is as other code in other applications making use of our libraries may break.  I suggest we rename it to `netsnmp_add_closable_notification_session` and then make `netsnmp_add_notification_session` a wrapper around that, and change all internal instances to the new version.

emilbart:
@hardaker: I updated the changes following your suggestion. Please consider updated commit.

hardaker:
Thanks; that looks better.

Issue #241, opened 2020-12-21
TLSUDP/TLSTCP: intermediate certificates not supported


minfrin:
When configuring TLS and specifying certificates provided either by a public CA or a private PKI that are deployed in a certificate hierarchy, the intermediate certificates are not passed to SSL_CTX.

As a result, intermediate certs are not sent to the other side during TLS handshake, and intermediate certs are not taken into account when verifying the peer. Certificate validation fails as a result.

Patch to follow.

Issue #242, opened 2020-12-21
netsnmp_cert_trust_ca: Unexpectedly wide trust behaviour


minfrin:
When we're asked to trust a certificate using the netsnmp_cert_trust_ca(), instead of trusting the certificate passed, we walk up any intermediate certificates and trust the deepest certificate in the chain.

This means if we're not careful, we would be silently trusting a vastly wider set of certificates than we otherwise intended.

This code currently has no effect, as net-snmp doesn't yet support intermediate certificates. As soon as intermediate certificates are supported, this code could kick in and start working.

https://github.com/net-snmp/net-snmp/blame/master/snmplib/cert_util.c#L2055

PR #244, opened 2020-12-22
Add support for intermediate certificates.

minfrin:
After constructing a certificate chain, pass the chain to openssl so the chain
is taken into account while negoiating SSL. Required to support certificates
generated by a PKI, or by a CA like Let's Encrypt.

With intermediates being supported, the "actually trust the topmost root
certificate instead of the certificate you specified" behaviour of
netsnmp_cert_trust_ca() is no longer used, and the certificate specified is
trusted instead.

Fixes https://github.com/net-snmp/net-snmp/issues/241 and
https://github.com/net-snmp/net-snmp/issues/242.


bvanassche:
Thanks for the patch. This patch has been applied on the v5.9 and master branches.

minfrin:
Can you confirm this PR can be closed?

bvanassche:
Confirmed.

PR #246, opened 2020-12-25
fix: MacOS Apple Silicon build

andreinechaev:


bvanassche:
The config.guess file comes from the automake project. Maintaining local changes for any files generated by autoconf, automake or libtool is painful. Has this patch already been submitted to the maintainers of this script? See also https://savannah.gnu.org/projects/config/.

andreinechaev:
Hello @bvanassche, sorry for the late replay. No, I haven't submitted any PRs to the project you mentioned. I will review the possibilities and the need. Thank you.

hardaker:
Yes, unfortunately we can't track changes to an upstream generated file easily.  The right thing to do is get it fixed there instead.

Issue #245, opened 2020-12-22
DTLSUDP: unable to load multiple certificates from a single PEM file


minfrin:
With PEM files, the traditional mechanism for specifying certificates is to group certificates together concatenated one after the other.

In the case of leaf certificates, the leaf certificate is specified, followed by intermediate certs, and an optional key.

In the case of CA certificates, the list of CA certs is typically published in a single file, example /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem .

In the case of net-snmp however, only the very first certificate in the file is taken into account, all other subsequent certificates are ignored. This makes net-snmp work differently to all the other openssl based software out there, and in turn makes it very difficult to configure, as it is not immediately obvious as to why the standard configuration used everywhere else doesn't work.

A further problem is that openssl doesn't provide the tools to split certificates into separate files. At best you can split certs into separate types (example leaf, intermediate, CA), but if you have two intermediate certs splitting the two requires custom code. All of this makes automation very difficult, and in turn makes it difficult to integrate net-snmp with Let's Encrypt.

To fix this, net-snmp needs to read multiple certs from each file, matching the behaviour of other openssl based software.

https://github.com/net-snmp/net-snmp/blob/ed90aaaaea0d9cc6c5c5533f1863bae598d3b820/snmplib/cert_util.c#L876

Patch to follow.

PR #250, opened 2020-12-30
Read intermediate certificates from PEM files

minfrin:
This patch lifts the restriction that a PEM file can only contain
one certificate. This brings net-snmp in line with the behaviour
of other applications that use PEM files.

Multiple certificates are read from each file, and the offset of
each certificate is added to the index for efficient retrieval.

The first certificate in the chain is eligible to be either an
identity certificate or a remote certificate. Intermediate
certificates are ignored for this purpose, maintaining existing
behaviour with respect to peerCert and remoteCert.

Trusted CA certificates are still required to be listed one per
file, this restriction will be lifted in a separate patch.

Fixes https://github.com/net-snmp/net-snmp/issues/245

### DIFF
--- a/include/net-snmp/library/cert_util.h
+++ b/include/net-snmp/library/cert_util.h
@@ -29,7 +29,7 @@ extern "C" {
     typedef struct netsnmp_cert_common_s {
         char           *dir;
         char           *filename;
-        
+
         u_char          type;
         u_char          allowed_uses;
         u_char          _pad[2]; /* for future use */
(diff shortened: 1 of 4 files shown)


minfrin:
Looks like this one was already committed in https://github.com/net-snmp/net-snmp/commit/916110b15f28f8ae04598ad637ac26e4585ff385

Can you confirm?

bvanassche:
Sure, confirmed.

PR #252, opened 2021-01-03
Allow trusting of more than one certificate per file

minfrin:
In a typical PEM encoded trust store, certificates are listed one
after the other in the same file. Lift the restriction that only
the first matching trusted certificate is trusted, and instead
trust all certificates that match a given PEM file.

This allows net-snmp to be used with standard PKI, such as Let's
Encrypt.

Fixes https://github.com/net-snmp/net-snmp/issues/245


hardaker:
Looks useful, thanks!

PR #257, opened 2021-01-05
Fix compilation with feature NETSNMP_FEATURE_REMOVE_LOGGING_STDIO

acafresna:
Fix compilation when feature NETSNMP_FEATURE_REMOVE_LOGGING_STDIO is used. Currently when this feature is set in include/net-snmp/net-snmp-config.h, according to INSTALL instructions, the program won't build. This fixes this problem.


hardaker:
Looks good, thanks!

PR #274, opened 2021-03-09
Enable systemd socket activation for UDP sockets

wferi:
For example if snmptrapd is UDP socket-activated, it still tries to bind the
socket passed in by systemd.  This is not permitted (the default port is "low",
thus privileged) and also invalid, because the port is already taken by systemd.

Since the code handling UDP is structured differently than TCP, that
method isn't directly applicable, but I tried to copy the idea as far as
possible. Other solutions are certainly possible if preferred.


wferi:
If the general direction is acceptable, I'm willing to extend this to the other transports as needed.

hardaker:
Looks like a reasonable solution to me.  thanks for the submission!

Issue #199, opened 2020-10-12
Parsing IP address twice causes Error: Adress family for hostname not supported


jridky:
"authCommunity" directive with an IP address as "SOURCE" parameter causes snmptrapd to generate an error log "Address family for hostname not supported", which is quite misleading.

How reproducible:
when an IP address or an IP network address ( IP/mask-length ) in a "authCommunity" directive in /etc/snmp/snmptrapd.conf
ex. authCommunity log,execute,net public 192.168.0.1

Steps to Reproduce:
1. create snmptrapd.conf with the following command.
```  # echo "authCommunity log,execute,net public 192.168.0.1" > /etc/snmp/snmptrapd.conf ```
2. start snmptrapd daemon.
```  # snmptrapd -Lf /tmp/s.log```

Actual results:
an error message appears in the log file "/tmp/s.log" like:
```
 /etc/snmp/snmptrapd.conf: line 1: Error: Address family for hostname not supported
 net-snmp: 1 error(s) in config file(s)
 NET-SNMP version 5.9
```

Expected results:
no errors should be reported.

Additional info:
This issue is caused by parsing IP address twice - first as IPv4 and as IPv6 at second, even if the address had proved to be a valid IPv4 address this error is probably generated by second attempt (IPv6 parsing).

Code reference:
https://github.com/net-snmp/net-snmp/blob/master/agent/mibgroup/mibII/vacm_conf.c#L953-L988

jridky:
@bvanassche any thoughts?

bvanassche:
Unfortunately I'm not familiar with the vacm_conf.c code.

jridky:
Patch with possible solution is ready. May I ask for review? @fenner @bvanassche

PR #276, opened 2021-03-12
Prevent parsing IP address twice (#199)

jridky:
This fixes issue, that is caused by parsing IP address twice. First as IPv4 and as IPv6 at second, even though the address was
properly parsed as a valid IPv4 address. See issue #199 for more details


bvanassche:
Thanks for the patch!

Issue #283, opened 2021-04-14
An error can be observed during parallel compilation of net-snmp.


JiangHeng12138:
when i use make -j32 to compile net-snmp，Sometimes the following errors occur：
```
[...]

PR #284, opened 2021-04-14
fix parallel compile error

JiangHeng12138:
Corresponding Issue: 
https://github.com/net-snmp/net-snmp/issues/283


kloczek:
BTW parallel. 
Parallel install target trashes installing libraries and usually agent library after all is with 0 bytes lenght.

bvanassche:
Both patches have been applied so I'm closing this pull request.

PR #296, opened 2021-05-27
Fix inconsistent use of #if vs. #ifdef NETSNMP_CAN_USE_SYSCTL

ThomasVie:
NETSNMP_CAN_USE_SYSCTL could be not defined.
Fix compile warning / error if compiled with the compile flags
-Wundef or -Werror=undef.


bvanassche:
This patch has been applied manually. Thanks for the patch!

PR #306, opened 2021-06-16
Perl agent: setValue() value conversion and logging improvements

guillemj:
This PR includes few fixes and improvements to the setValue() value conversion by using a scaled epsilon to check for rounding relative errors, and logging more useful double values, and adding missing newlines.


bvanassche:
Thanks for the patches!

Issue #312, opened 2021-07-02
SIGHUP handling of different persistent configs.


EZenderink:
Hi,

I am currently working on a solution that allows us to perform a soft restart (SIGHUP) of net-snmp while swapping the persistent configuration files. Originally we used the same implementation used in `snmpd.c` for SIGHUP but added function calls to allow different persistent configurations during restart to work with VACM/USM, as well as being able to reconfigure the port the agent listens on (code not shown here):

    void RestartSnmp(void)
    {
        // Shut down master agent to listen on different udp ports
        shutdown_master_agent();

        // Call update config callbacks.
        snmp_call_callbacks(SNMP_CALLBACK_APPLICATION, SNMPD_CALLBACK_PRE_UPDATE_CONFIG, NULL);

        // Store persistent data
        snmp_store("a-config");

        // Free stored configuration.
        free_config();
        
        // unregister ipAddressTable
        // Prevent stderr warning:
        // duplicate registration: MIB modules ipAddressTable and ipAddressTable
        // (oid .1.3.6.1.2.1.4.34).
        unregister_mib((oid[]){1, 3, 6, 1, 2, 1, 4, 34}, 8);

        // Initialize configuration tokens for usm and vacm
        // fix USM tokens (createUser, etc) unknown warning
        init_usm_conf("b-config");

        // fix VACM tokens (rwuser, vacmView, vacmAccess etc...) unknown warning
        init_vacm_conf();

        // Read configurations (aka baseconfig)
        read_configs();

        // Read new persistent config (aka non legacy or legacy)
        init_agent_read_config("b-config");

        // Start master agent to listen on different udp ports
        if (init_master_agent())
        {
            return 0;
        }
        return 1;
    }

As can be read in the comments in the snippet above I had to unregister the ipAddressTable OID to prevent the "duplicate registration" warning. To be able to instantiate the USM users and prevent the `unknown token` warnings for USM configuration within the persistent `b-config.conf` we had to call `init_usm_conf()`  giving the application name matching the persistent config that we want net-snmp to read/write to. 

This causes however the problem that every time this is called the callback for USM for SNMP_CALLBACK_STORE_DATA is registered, resulting in multiple duplicate entries in the persistent config files when performing soft restarts. By unregistering this callback before registering the callback in `init_usm_conf`, the issue went away for USM. However this does not seem  like a suitable fix. 

[...]

PR #313, opened 2021-07-05
#312 duplicate callbacks fix

EZenderink:
See ticket #312 for more information.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,7 +3,7 @@ image:
   - Ubuntu1604
   - Ubuntu2004
   # - macos
-  - macos-mojave
+  # - macos-mojave
   - Visual Studio 2019
 
 environment:
--- a/ci/before.sh
+++ b/ci/before.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Instead of relying on the hosts file provided by the CI host, replace it.
 # See also
@@ -7,8 +7,20 @@ sudo sh -c 'printf "127.0.0.1 ipv4-loopback\n::1 localhost ipv6-localhost ipv6-l
 
 case "$(uname)" in
     Linux)
-	sudo sh -c 'apt-get install -y python3-dev'
-	sudo sh -c 'apt-get install -y libmariadbclient-dev || sudo apt-get install -y libmariadb-client-lgpl-dev'
+	packages=(
+	    libatm1-dev
+	    libncurses5-dev
+	    libperl-dev
+	    libsensors-dev
+	    libsensors4-dev
+	    libssh2-1-dev
+	    libssl-dev
+	    python3-dev
+	)
+	for p in "${packages[@]}"; do
+	    sudo sh -c "apt-get install -y $p"
+	done
+	sudo sh -c 'apt-get install -y libmariadbclient-dev || sudo apt-get install -y libmariadb-client-lgpl-dev || sudo apt-get install -y libmysqlclient-dev'
 
 	# Add an IPv6 config - see the corresponding Travis issue
 	# https://github.com/travis-ci/travis-ci/issues/8361


bvanassche:
This patch breaks netsnmp_register_callback() since its documented behavior is as follows: `Handler invocation priority. When multiple handlers have been registered for the same (major, minor) callback event type, handlers with the numerically lowest priority will be invoked first. Handlers with identical priority are invoked in the order they have been registered.`

EZenderink:
Hi, so if I understand correctly, the fix has to be done within each mib module (ifxtable/iftable, usm, serialno, etc) to prevent having multiple SNMP_CALLBACK_STORE_DATA callbacks registered which in our scenario is causing the persistent config file to grow with duplicate entries for serialno, ifXtable and usm users, everytime the mib modules are reinitialized (which we have to do since we want to reload a different persistent config on sighup). Should I create a pull request with changes so that these modules unregister the callback on shutdown (as I would expect them to)? Or is the way we try to reload net-snmp not the correct way/intended way? (Which if that is the case, is it at all possible to reload net-snmp using sighup  while providing a completely different persistent config? This is a big use case within our custom agent.)

bvanassche:
There is probably an issue with how Net-SNMP is reloaded. Has it been considered to send SIGHUP to snmpd such that the code in snmpd.c for reloading a configuration is triggered? From snmpd.c:
~~~
static void
snmpd_reconfig(void)
{
#ifdef HAVE_SIGPROCMASK
    sigset_t set;
    int ret;

    sigemptyset(&set);
    sigaddset(&set, SIGHUP);
    ret = sigprocmask(SIG_BLOCK, &set, NULL);
    netsnmp_assert(ret == 0);
#endif
    reconfig = 0;
    snmp_log(LOG_INFO, "Reconfiguring daemon\n");
    /* Stop and restart logging.  This allows logfiles to be rotated etc. */
    netsnmp_logging_restart();
    snmp_log(LOG_INFO, "NET-SNMP version %s restarted\n",
             netsnmp_get_version());
    update_config();
    send_easy_trap(SNMP_TRAP_ENTERPRISESPECIFIC, 3);
#ifdef HAVE_SIGPROCMASK
    ret = sigprocmask(SIG_UNBLOCK, &set, NULL);
    netsnmp_assert(ret == 0);
#endif
}
~~~
As far as I can tell the above code does not trigger reregistration of any callbacks.

EZenderink:
@bvanassche 

Thank you for your reply. Initially our code started out with that implementation. However we have the following use cases which required some alteration for us to get it working (with the mentioned issues that followed):

Scenario: 
- The following configurations are stored in the persistent config dir: a-config.conf, b-config.conf
- Start net-snmp with a-config.conf
- On sighup: store created usm users in a-config.conf
- On sighup: reload netsnmp with a b-config.conf persistent configuration file (swapping user usm/vacm, etc...)
   - This also means: clearing all currently stored usm/vacm configuration from a-config.conf, before loading in the different persistent config with different usm/vacm configuration.
- On sighup: be able to restart the SNMP agent to listen on different UDP port.


Basically it comes down to restarting net-snmp as if the full agent restarts by killing the application and restarting, but then triggered by SIGHUP without actually killing the application (we actually tried this at first using snmp_init and shutdown as you normally would. However this caused significant memory leaks unless the application actual exits after starting the shutdown procedure. ) I take it that net-snmp currently does not support this functionality, and expects the same config to be reloaded after sighup, correct?

In the first use case, the function `update_config()`  updates the persistent config as expected. However, we cannot change to which persistent config during restart. 

So far I could tell there was no other way to restart snmp with a different persistent configuration than calling `init_agent_read_config(const char *)` and providing the file name of the persistent configuration (without the extension) as argument.

Further more, to get USM to work with a different persistent configuration, I had to call 'shutdown_usm()' and then 'init_usm_conf(const char *app)' which is also the file name of the persistent configuration (without the extension). I would perhaps have expected shutdown_usm to unregister the SNMP_CALLBACK_STORE_DATA callback which init_usm_conf registers. Currently every time `init_usm_conf(const char *app)` is called, it registers this callback:

    void
    init_usm_conf(const char *app)
    {
        register_config_handler(app, "usmUser",
                                      usm_parse_config_usmUser, NULL, NULL);
        register_config_handler(app, "createUser",
                                      usm_parse_create_usmUser, NULL,
                                      "username [-e ENGINEID] (MD5|SHA|SHA-512|SHA-384|SHA-256|SHA-224|default) authpassphrase [(DES|AES|default) [privpassphrase]]");

        /*
         * we need to be called back later
         */
        snmp_register_callback(SNMP_CALLBACK_LIBRARY, SNMP_CALLBACK_STORE_DATA,
                               usm_store_users, NULL);
    }

As can be seen it requires the app name to know for which configuration it needs to load the users from (as far as I understood). Without calling this function after a sighup when loading in a different persistent config, it does not initialize the users stored in the persistent config nor does it store newly created users to the different persistent config. 

Thus basically it boils down to: on sighup I would like to be able fully reconfigure net-snmp, using different persistent config files.  Currently it is USM that is blocking this functionality. 

The current way of how I reinitialize netsnmp with different configurations during sighup in full is shown below:

    int RestartNetSnmpSighup(char *aConfPersistent)
    {
        // Shut down current instance to allow for reconfiguration of UDP
        shutdown_master_agent();

        // Call update config callbacks (see update_config() in agent_read_config.h).
        snmp_call_callbacks(SNMP_CALLBACK_APPLICATION, SNMPD_CALLBACK_PRE_UPDATE_CONFIG, NULL);

        // Store the currently configured persistent data to the active persistent configuration file.
        snmp_store(SnmpAgent_confPersistent(NULL));

        // Free stored configuration within netsnmp lib. (see update_config() in agent_read_config.h).
        free_config();

        // Shutdown USM instance.
        shutdown_usm();

        // Unregister the token for this specific configuration
        // @note, this is to prevent tokens being handled for the wrong application
        // instance, might not be necessary though.
        unregister_config_handler(SnmpAgent_confPersistent(NULL), "createUser");
        unregister_config_handler(SnmpAgent_confPersistent(NULL), "usmUser");

        // To fix the Warning: Unknown token: ifXtable, it is necessary
        // to shut down the inferace, and restart the interface once the
        // new persistent config file name is set.
        // fix Warning: Unknown token: ifXTable.
        _ifXTable_shutdown_interface(ifXTable_registration_get());
        // shutdown_ifXTable();

        // restart logging
        netsnmp_logging_restart();

        // Make sure that the correct configs are read when restarting SNMP
        // This appends the extension ".active" the to be inactive configuration file
        // So that netsnmp does not read this config file.
        SnmpAgent_activateConf(aConfPersistent, SnmpAgent_confPersistent(NULL));

        // This function statically stores which persistent configruation is active
        // by calling SnmpAgent_confPersistent(NULL) a char * is returned containing
        // the name of the currently active persistent config file.
        SnmpAgent_confPersistent(aConfPersistent);

        // unregister ipAddressTable
        // Prevent stderr warning:
        // duplicate registration: MIB modules ipAddressTable and ipAddressTable
        // (oid .1.3.6.1.2.1.4.34).
        unregister_mib((oid[]){1, 3, 6, 1, 2, 1, 4, 34}, 8);

        // Unregister ifXtable to allow reregistration of the table(see below)
        unregister_mib((oid[]){1, 3, 6, 1, 2, 1, 31, 1, 1}, 9);

        // Unregister snmpSetSerialNo (oid .1.3.6.1.6.3.1.1.6.1) to allow reregistration of the table
        // (see below)
        unregister_mib((oid[]){1, 3, 6, 1, 6, 3, 1, 1, 6, 1}, 10);

        // Read new persistent config (aka non legacy or legacy)
        init_agent_read_config(SnmpAgent_confPersistent(NULL));

        // Initialize configuration tokens for usm and vacm
        // fix createUser token unknown warning
        // without shutting down USM and reinitializing USM it does not apply 
        // usm configuration from the newly loaded persistent configuration, and
        // generates Unknown token warnings on stderr for all usm related configurations
        //
        // Note: IMPORTANT: this also registers the SNMP_CALLBACK_STORE_DATA
        // callback again, resulting in duplicate entries in the persistent config!
        init_usm_conf(SnmpAgent_confPersistent(NULL));

        // fix rwuser token unknown warning
        // without re-initializing vacm the warning unkonwon token for all vacm
        // related configs is printed to console and the vacm entries in the 
        // persistent config file is not being configured.
        init_vacm_conf();

        // fix Warning: Unknown token: ifXTable.
        // By shutting down the ifXtable interface and restarting the ifXtable
        // interface we are aable to negate the ifXtable unknown token warning.
        //
        // Note: IMPORTANT: this also registers the SNMP_CALLBACK_STORE_DATA
        // callback again, resulting in duplicate entries in the persistent config!
        _ifXTable_initialize_interface(ifXTable_registration_get(), 0);

        // fix Warning: Unknown token: setserialno.
        // without re-initializing setSerialNo the warning unkonwon token for setserialno
        // is printed to stderr console and might not be configured
        //
        // Note: IMPORTANT: this also registers the SNMP_CALLBACK_STORE_DATA
        // callback again, resulting in duplicate entries in the persistent config!
        init_setSerialNo();

        // Read configurations (see update_config() in agent_read_config.h).
        read_configs();

        // Start master again to allow for reconfiguration of UDP
        if (init_master_agent())
        {
            printf("Failed to restart master agent!");
            return 1;
        }
        return 0;
    }

With my pull request ignoring the registration of an already registered callback, it fixes the issues I have noted in the "IMPORTANT" comments in the code above.

Obviously the way ifXtable and serialno are being reinitialized should not be necessary at all, but without doing so, the tokens in the persistent configuration are not loaded in properly and generate warnings such as "unknown token" on stderr. Perhaps more modules face the same issue but these are the ones we came across while trying to implement this functionality. 

Perhaps this is way out of scope for how net-snmp currently works, but it is quit important to us to get this working.

PR #317, opened 2021-07-12
Fix compiler warnings

ncopa:


bvanassche:
A slightly modified version of these patches has been applied on the v5.9 and master branches. Thanks for the patches!

PR #324, opened 2021-08-04
Add ignoremount config option

po5857:
We monitor many hosts, but we want the ability to ignore certain mounts for a couple of reasons:

1) some mounts (such as backup mounts) will always be near 100%, but we do not care
2) some mounts are on systems which get so heavily used that snmpd will hang trying to check on their usage.  While this is obviously a problem we should solve (fix the storage system), having snmpd ignore them is a cheaper option

The attached patch adds an "ignoremount" option to snmpd.conf which can be used like so:
```
ignoremount /disk1
ignoremount /disk2
```
This makes the snmpd agent completely skip over the configured mounts for all mount related inquiries.  We have been dragging this patch along for a few years now, and would love to see it in upstream.

### DIFF
--- a/agent/mibgroup/hardware/fsys/fsys_mntent.c
+++ b/agent/mibgroup/hardware/fsys/fsys_mntent.c
@@ -59,6 +59,9 @@
 
 #endif
 
+static void     parse_mount_config(const char *, char *);
+static void     free_mount_config(void);
+
 /*
  * File systems to monitor and that are not covered by any hrFSTypes
  * enumeration.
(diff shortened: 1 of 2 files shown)


agent/mibgroup/hardware/fsys/fsys_mntent.c
```diff
@@ -176,7 +179,240 @@ _fsys_type( char *typename )
 void
 netsnmp_fsys_arch_init( void )
 {
-    return;
+    snmpd_register_config_handler("ignoremount", parse_mount_config,
+                                  free_mount_config, "name");
```
bvanassche, line 183: This is the wrong location to add an snmpd_register_config_handler() call. The file in which this call has been added, fsys_mntent.c, implements architecture support for Linux but it not used on e.g. FreeBSD systems. Please move this call into init_hw_fsys().

agent/mibgroup/hardware/fsys/fsys_mntent.c
```diff
@@ -176,7 +179,240 @@ _fsys_type( char *typename )
...
+    char           *name = NULL, *p = NULL, *d_str = NULL, c;
+    unsigned int    i, neg, c1, c2;
+    char           *st = NULL;
+
+    name = strtok_r(cptr, " \t", &st);
+    if (!name) {
+        config_perror("Missing NAME parameter");
+        return;
+    }
+    d_new = (conf_mount_list *) malloc(sizeof(conf_mount_list));
```
bvanassche, line 223: Please do not cast the malloc() return value. Such casts are not necessary since the return type of malloc() is `void *`.

agent/mibgroup/hardware/fsys/fsys_mntent.c
```diff
@@ -176,7 +179,240 @@ _fsys_type( char *typename )
...
+                neg = 0;
+            }
+            while (*name && *name != ']') {
+                c1 = ((unsigned int) *name++) & 0xff;
+                if (*name == '-' && *(name + 1) != ']') {
+                    name++;
+                    c2 = ((unsigned int) *name++) & 0xff;
+                } else {
+                    c2 = c1;
+                }
```
bvanassche, line 270: Please store the entire regular expression as a string instead of storing it in a data structure.

agent/mibgroup/hardware/fsys/fsys_mntent.c
```diff
@@ -176,7 +179,240 @@ _fsys_type( char *typename )
...
+                result = 1;
+            }
+            break;
+        case ITEM_ANY:
+            if (*name)
+                result = match_mount_config_item(name + 1,
+                                                di_ptr->item_next);
+            break;
+        }
+    } else {
```
bvanassche, line 393: Please use libpcre instead of implementing regular expression matching from scratch. See also http://www.pcre.org/. See also the Net-SNMP --with_pcre configure option.

bvanassche (changes requested):

po5857:
Your first 2 points make sense, but on your second 2 points please note that my handling of ignoremount was a shameless copy from ignoredisk handling via parse_disk_config in agent/mibgroup/host/hr_disk.c.  Are the changes you suggest also needed there?

bvanassche:
> Your first 2 points make sense, but on your second 2 points please note that my handling of ignoremount was a shameless copy from ignoredisk handling via parse_disk_config in agent/mibgroup/host/hr_disk.c. Are the changes you suggest also needed there?

Moving the registration of the config handler can wait until later (after this commit went in).

Using libpcre instead of adding a new regular expression matcher is something I care about since the new regular expression matcher consists of a considerable number of codes and hence would increase the code maintenance workload.

bvanassche:
Please reopen this pull request after the review comments have been addressed.

PR #327, opened 2021-08-05
Add config parser fuzzer

DavidKorczynski:
Add config parser fuzzer


bvanassche:
Thanks for the patch.

Regarding the fuzzers, how about moving the fuzzing/ directory to testing/fuzzing/ ? The latter is probably a better location for the directory with fuzzers? If you agree, can you submit a pull request that realizes this change?

PR #328, opened 2021-08-06
Fuzzing: move folder to testing

DavidKorczynski:
Fuzzing: move folder to testing as suggested here https://github.com/net-snmp/net-snmp/pull/327#issuecomment-893957724


bvanassche:
Thanks!

PR #329, opened 2021-08-07
Fuzzing: Add config memory read fuzzer

DavidKorczynski:
Add fuzzer that target read_config_read_memory


bvanassche:
Thanks for the patch!

PR #333, opened 2021-08-17
fuzzing: add new api fuzzer

DavidKorczynski:
@bvanassche this PR adds a new API fuzzer. The fuzzer relies on a small library that makes fuzzer-writing easier by taking care of extracting various data types, e.g. null-terminated strings, seeded with fuzzer input. The library is available here: https://github.com/AdaLogics/fuzz-headers and I use it to fuzz other important OSS projects, e.g. Apache-httpd, and soon also OpenVPN and more. I would be very happy to continue using that library for writing net-snmp fuzzers. However, I see that you are refining the infrastructure around building fuzzers so am not entirely sure where I should include downloading the small library (it's just a header file for now). Could you perhaps elaborate a bit on how you prefer to have the build scripts? I could also help with migrating the build scripts to the OSS-Fuzz infra as I am very familiar with that.

### DIFF
--- a/include/net-snmp/ada_fuzz_header.h
+++ b/include/net-snmp/ada_fuzz_header.h
@@ -1,0 +1,119 @@
+/*
+  * Copyright (c) 2021, Net-snmp authors
+  * All rights reserved.
+  *
+  * Redistribution and use in source and binary forms, with or without
+  * modification, are permitted provided that the following conditions are met:
+  *
+  * * Redistributions of source code must retain the above copyright notice, this
+  *   list of conditions and the following disclaimer.
+  *
+  * * Redistributions in binary form must reproduce the above copyright notice,
+  *   this list of conditions and the following disclaimer in the documentation
+  *   and/or other materials provided with the distribution.
+  *
+  * * Neither the name of the copyright holder nor the names of its
+  *   contributors may be used to endorse or promote products derived from
+  *   this software without specific prior written permission.
+  *
+  * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+  * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+  * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+  * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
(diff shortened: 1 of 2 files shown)


bvanassche:
How about adding source files that are only used by the fuzz tests somewhere under testing/fuzzing/?

My preference is to have the build script for the fuzz tests inside the Net-SNMP tree instead of inside the oss-fuzz tree. Having the build script inside the Net-SNMP has the following advantages:
* This makes it easy to build the fuzz tests while building the rest of the Net-SNMP code.
* This makes it easy to analyze the fuzz tests with Coverity (Coverity has already detected a bug in these tests).

The following pull request shows where I would like to go (does not yet build): https://github.com/google/oss-fuzz/pull/6231.

DavidKorczynski:
Sounds good with the build script although at the moment I find it slightly confusing since we have multiple build scripts `build-fuzz-tests.sh` and `build.sh`. In either case I am happy for complete this set up and then I will adapt accordingly. Unifying this would make things easier.

However, about the question of a library above: it's a library used to write fuzzers so I don't think we should add the source files of that library here. My question is more in which script would you prefer for me to download the library? I can do it in `build-fuzz-tests.sh`?

bvanassche:
This pull request breaks the build:
~~~
$ ci/build.sh
[ ... ]
making all in /usr/local/google/home/bvanassche/software/net-snmp/testing
testing/fuzzing/snmp_api_fuzzer.c:38:10: fatal error: 'ada_fuzz_header.h' file
      not found
#include "ada_fuzz_header.h"
         ^~~~~~~~~~~~~~~~~~~
1 error generated.
~~~

DavidKorczynski:
> This pull request breaks the build:
> 
> ```
> $ ci/build.sh
> [ ... ]
> making all in /usr/local/google/home/bvanassche/software/net-snmp/testing
> testing/fuzzing/snmp_api_fuzzer.c:38:10: fatal error: 'ada_fuzz_header.h' file
>       not found
> #include "ada_fuzz_header.h"
>          ^~~~~~~~~~~~~~~~~~~
> 1 error generated.
> ```

Correct: I mentioned this in the first comment in this PR *The fuzzer relies on a small library that makes fuzzer-writing easier by taking care of extracting various data types, e.g. null-terminated strings, seeded with fuzzer input. The library is available here: https://github.com/AdaLogics/fuzz-headers`* - those are my questions about where you would prefer for me to include the downloading of a fuzz helper library.

DavidKorczynski:
@bvanassche now I just included the header file directly in the includes. This compiles now and it runs well with OSS-Fuzz!

bvanassche:
Thanks, but please take a look at commit 05a463fd4ca6ffa51698e00ddd6e30d0db1b5efd.

PR #334, opened 2021-08-18
Deprecated retranstime syscall fix

EZenderink:
Hi,

I ran into an issue that caused the following warning(s):

`localhost kernel: ICMPv6: process 'snmp' is using deprecated sysctl (syscall) net.ipv6.neigh.bond0.retrans_time - use net.ipv6.neigh.bond0.retrans_time_ms instead`

 In our case, the snmp agent ran in a seperate network namespace, which does not have access to the "/proc/sys/net/ipv%d/neigh/default" network interface, which is currently used to determine if "reachable_time_ms" exists. By moving this check to where the actual "reachable_time_ms" value is read from a _specific_ network interface. This issue is solved. 

I noticed the same issue for "base_reachable_time" hence I fixed that in this pull request as well. 

Kind Regards,

Eldin Zenderink

### DIFF
--- a/ci/net-snmp-configure
+++ b/ci/net-snmp-configure
@@ -137,7 +137,7 @@ options=()
 options+=(--enable-developer)
 options+=(--enable-ipv6)
 options+=("--prefix=/usr/local/net-snmp-${branch_name}")
-options+=("--with-cflags=$cflags")
+options+=("--with-cflags=$CFLAGS")
 options+=(--with-defaults)
 transports=()
 # Mib names can be found as follows:


bvanassche:
Please combine the two patches into a single patch, e.g. with `git rebase -i` followed by `git push ${remote} +${branch_name}`.

EZenderink:
@bvanassche Sorry about this, I managed to make a bit of  a mess of it, and accidently force push a commit which closed the pull request. I have managed to combine the changes into one commit, should a I create a new pull request or are you able to reopen this pull request on your end?

PR #335, opened 2021-08-19
Fix deprecated retrans_time and base_reachable_time syscall warning

EZenderink:
Hi,

I ran into an issue that caused the following warning(s):

localhost kernel: ICMPv6: process 'snmp' is using deprecated sysctl (syscall) net.ipv6.neigh.bond0.retrans_time - use net.ipv6.neigh.bond0.retrans_time_ms instead

In our case, the snmp agent ran in a seperate network namespace, which does not have access to the "/proc/sys/net/ipv%d/neigh/default" network interface, which is currently used to determine if "reachable_time_ms" exists. By moving this check to where the actual "reachable_time_ms" value is read from a specific network interface. This issue is solved.

I noticed the same issue for "base_reachable_time" hence I fixed that in this pull request as well.

Kind Regards,

Eldin Zenderink

### DIFF
--- a/agent/mibgroup/if-mib/data_access/interface_linux.c
+++ b/agent/mibgroup/if-mib/data_access/interface_linux.c
@@ -183,36 +183,6 @@ static void init_libpci(void)
 void
 netsnmp_arch_interface_init(void)
 {
-    /*
-     * Check which retransmit time interface is available
-     */
-    char proc_path[ 64+IF_NAMESIZE];
-    char proc_path2[64+IF_NAMESIZE];
-    struct stat st;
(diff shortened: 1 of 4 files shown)


EZenderink:
Hi @bvanassche, perhaps did I mess something up in this pull request? Please let me know if there still is something missing before this pull request can be integrated. 

Kind Regards,

Eldin Zenderink

bvanassche:
A modified version of the provided patches has been checked in as commit 85bff70932178ccd08b55112cadcb0e1bedb909d. Please take a look.

Issue #46, opened 2019-12-02
Memleak netsnmp_


EZenderink:
Hi,

I found the following memory leak. 

    Direct leak of 68 byte(s) in 4 object(s) allocated from:
        #0 0x7fbe42082330 in __interceptor_malloc (/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
        #1 0x564430bf1b5a in netsnmp_memdup /home/Code/packages/net-snmp/core/snmplib/tools.c:288

    SUMMARY: AddressSanitizer: 68 byte(s) leaked in 4 allocation(s).

This is obviously not enough to know why it happens, so I followed our steps on why it happens.

Basically, we enter a function within `agent_trap.c` called `netsnmp_create_v3user_notification_session`.  In here memory is duplicated at line 1622:

        /** engineId */
        session.contextEngineID = netsnmp_memdup(usmUser->engineID,
                                             usmUser->engineIDLen);
        session.contextEngineIDLen = usmUser->engineIDLen;

However at some point we enter the `goto bail;` part and the following is performed, at line 1662:

      bail:
        /** free any allocated mem in session */
        SNMP_FREE(session.securityAuthProto);
        SNMP_FREE(session.securityPrivProto);

When adding `SNMP_FREE(session.contextEngineID);` , changing it to: 

      bail:
        /** free any allocated mem in session */
        SNMP_FREE(session.securityAuthProto);
        SNMP_FREE(session.securityPrivProto);
        SNMP_FREE(session.contextEngineID);

The memory leak is gone.

Issue #16, opened 2019-09-26
Memory leak when using localized keys


EZenderink:
Hi,

When performing a localized key change our sanitizer noticed the following memory leak:

```

=================================================================
==16942==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 128 byte(s) in 1 object(s) allocated from:
    #0 0x7f26cddb3330 in __interceptor_malloc (/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x5603766d2860 in usmDHGetUserKeyChange snmp-usm-dh-objects-mib/usmDHUserKeyTable/usmDHUserKeyTable_data_get.c:78
    #2 0x560376746dc3  (/usr/local/bin/snmp+0x26ddc3)

Direct leak of 128 byte(s) in 1 object(s) allocated from:
    #0 0x7f26cddb3330 in __interceptor_malloc (/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x5603766d3536 in usmDHSetKey snmp-usm-dh-objects-mib/usmDHUserKeyTable/usmDHUserKeyTable_data_set.c:77

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #0 0x7f26cddb3330 in __interceptor_malloc (/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x7f26cdacc328 in CRYPTO_zalloc (/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x192328)

Indirect leak of 128 byte(s) in 1 object(s) allocated from:
    #0 0x7f26cddb3330 in __interceptor_malloc (/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x7f26cdacc328 in CRYPTO_zalloc (/lib/x86_64-linux-gnu/libcrypto.so.1.1+0x192328)

SUMMARY: AddressSanitizer: 408 byte(s) leaked in 4 allocation(s).
```

This is tested with commit: [e8f2c077b81435099af1269eb40d5eb188ed0f28](https://github.com/net-snmp/net-snmp/tree/e8f2c077b81435099af1269eb40d5eb188ed0f28)

EZenderink:
[...]

PR #346, opened 2021-09-27
Fix memory leaks in USM and DH algorithm

EZenderink:
Hereby a patch for memory leaks we found during our implementation for usm and localized keys, reported some time ago (#16 and #46 ).

Please let us know if these changes are acceptable.


bvanassche:
A slightly modified patch has been applied on the v5.9 and master branches. Thanks for the patch!

PR #347, opened 2021-10-04
Avoid openssl error ring memory segmentation fault (NETSNMP C API)

lukas-wimmer:
There is an invalid double OPENSSL_free in the openssl at the same memory address caused by snmpTLSBaseDomain.c.
(Target memory is used by openssl to manage an internal error ring).

Due to the release of the protected openssl memory in the netsnmp library it can happen that the memory is released twice at a later point in time and a crash occurs.

We found the problem especially when multiple TLS-based traps are sent over the NETSMP C API and the trap target is not reachable. 

code in openSSL(1.1.1l)
https://github.com/openssl/openssl/blob/master/crypto/err/err.c
```
#define err_clear_data(p, i) \
do { \
if ((p)->err_data_flags[i] & ERR_TXT_MALLOCED) {\
OPENSSL_free((p)->err_data[i]); \
(p)->err_data[i] = NULL; \
} \
(p)->err_data_flags[i] = 0; \
} while (0)
```

### DIFF
--- a/snmplib/transports/snmpTLSBaseDomain.c
+++ b/snmplib/transports/snmpTLSBaseDomain.c
@@ -1228,16 +1228,9 @@ void _openssl_log_error(int rc, SSL *con, const char *location) {
         /* if we have a text translation: */
         if (data && (flags & ERR_TXT_STRING)) {
             snmp_log(LOG_ERR, "  Textual Error: %s\n", data);
-            /*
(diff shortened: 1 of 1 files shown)


bvanassche:
The two patches in the pull request have been combined and have been applied on the v5.9 and master branches. Thanks for the patches!

PR #353, opened 2021-10-14
snmpd: Fix memory allocation

DavidKorczynski:
Ensure memory allocation of more than 1 byte in case asp->vbcount is 0
Further clarification of this is found here:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39901#c1

This also fixes
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=39930

### DIFF
--- a/agent/snmp_agent.c
+++ b/agent/snmp_agent.c
@@ -3771,7 +3771,9 @@ handle_pdu(netsnmp_agent_session *asp)
     case SNMP_MSG_INTERNAL_SET_RESERVE1:
 #endif /* NETSNMP_NO_WRITE_SUPPORT */
         asp->vbcount = count_varbinds(asp->pdu->variables);
-        asp->requests = calloc(asp->vbcount, sizeof(netsnmp_request_info));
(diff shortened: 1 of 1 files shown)


agent/snmp_agent.c
```diff
@@ -3771,7 +3771,7 @@ handle_pdu(netsnmp_agent_session *asp)
     case SNMP_MSG_INTERNAL_SET_RESERVE1:
 #endif /* NETSNMP_NO_WRITE_SUPPORT */
         asp->vbcount = count_varbinds(asp->pdu->variables);
-        asp->requests = calloc(asp->vbcount, sizeof(netsnmp_request_info));
+        asp->requests = calloc(asp->vbcount+1, sizeof(netsnmp_request_info));
```
bvanassche, line 3774: If memory is allocated for one varbind more than needed, that will prevent Valgrind from detecting certain out-of-bounds errors. Can `asp->vbcount + 1` be changed into `asp->vbcount ? asp->vbcount : 1`?

PR #352, opened 2021-10-14
TSM with TLS: Support Subject Alternative with IP 

lukas-wimmer:
Currently SAN with IP is supported only in client certificates. (Verified by the agent)

Now server side certificates can also be verified with IP using Subject Alternative Names. 

According to rfc6353, Common Name verification is deprecated. 
So I am not sure if we should also set the X509_CHECK_FLAG_NEVER_CHECK_SUBJECT flag?
... but it would change the current behavior

For more details about used flags:
https://www.openssl.org/docs/man1.1.0/man3/X509_check_host.html

PR #355, opened 2021-10-19
[TSM] TLSTCP supports SAN IP in Server Certificate

lukas-wimmer:
Currently SAN with IP is supported only in client certificates. (Verified by the agent)

Now server side certificates can also be verified with IP using Subject Alternative Names.

Recreate Pull Request from 
https://github.com/net-snmp/net-snmp/pull/352

Fix Compile Error


bvanassche:
This patch has been applied manually on the master branch. Thanks for the patch!

PR #356, opened 2021-10-19
libsnmp: Fix a read-past-end-of-buffer

DavidKorczynski:
Prevent that cp2 reads past end of the cp buffer. This patch fixes
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36874


bvanassche:
Thanks for the patch!

PR #357, opened 2021-10-19
libsnmp: Fix spurious buffer read

DavidKorczynski:
Ensure 'ix' is in range of 'buf' so the code 'buf[ix] |= bit;' is valid.
This fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=37564


bvanassche:
Thanks for the patch!

PR #359, opened 2021-10-20
libnsmp: Fix off-by-one read

DavidKorczynski:
Prevent that the code 'line[linelen - 1]' reads a negative index.
This fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36908


bvanassche:
Thanks for the patch!

PR #369, opened 2021-11-01
libsnmp: Fix stack-based buffer overflow

DavidKorczynski:
Ensure len is above 0 before the code 'utc[len - 1]'.
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40508


bvanassche:
Thanks for the patch!

PR #370, opened 2021-11-01
snmpd: Ensure contextNameLen follows buffer size

DavidKorczynski:
Ensure strlen is equal to community_len before the code
  ```
  pdu->contextName    = strdup((char *) pdu->community);
  pdu->contextNameLen = pdu->community_len;
  ```

This ensures `contextNameLen` is not above the size of the buffer returned by `strdup`.
Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40431

### DIFF
--- a/agent/mibgroup/agentx/protocol.c
+++ b/agent/mibgroup/agentx/protocol.c
@@ -1643,6 +1643,11 @@ agentx_parse(netsnmp_session * session, netsnmp_pdu *pdu, u_char * data,
          * expects to find the context in the PDU's context field.  Therefore we
          * need to copy the context into the PDU's context fields.  */
         if (pdu->community_len > 0 && pdu->contextName == NULL) {
+            /*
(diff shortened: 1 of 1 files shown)


agent/mibgroup/agentx/protocol.c
```diff
@@ -1643,6 +1643,8 @@ agentx_parse(netsnmp_session * session, netsnmp_pdu *pdu, u_char * data,
          * expects to find the context in the PDU's context field.  Therefore we
          * need to copy the context into the PDU's context fields.  */
         if (pdu->community_len > 0 && pdu->contextName == NULL) {
+            if (strlen(pdu->community) != pdu->community_len)
```
bvanassche, line 1646: Please add a comment that explains that using strlen() is safe because snmp_clone_mem() '\0'-terminates its output or use memchr() instead of strlen().

bvanassche:
Thanks for the patch!

PR #371, opened 2021-11-01
testing/fuzzing: Declare functions in fuzz helper

DavidKorczynski:
Declare functions in fuzz helper to remove warnings during build.


bvanassche:
One change per patch please for future patches. Anyway, thanks for the patch.

PR #372, opened 2021-11-01
testing/fuzzing: Add transport fuzzer

DavidKorczynski:
Add transport fuzzer which was previously removed because it used
undeclared functions (https://github.com/net-snmp/net-snmp/pull/366).


bvanassche:
Thanks for the patch!

PR #373, opened 2021-11-05
testing/fuzzing: Fix cast to avoid compiler warning

DavidKorczynski:

### DIFF
--- a/testing/fuzzing/agentx_parse_fuzzer.c
+++ b/testing/fuzzing/agentx_parse_fuzzer.c
@@ -53,7 +53,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     netsnmp_session session;
 
     session.version = AGENTX_VERSION_1;
-    if (agentx_parse(&session, pdu, (unsigned char *)data, size) == SNMP_ERR_NOERROR) {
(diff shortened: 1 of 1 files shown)


testing/fuzzing/agentx_parse_fuzzer.c
```diff
@@ -53,7 +53,8 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     netsnmp_session session;
 
     session.version = AGENTX_VERSION_1;
-    if (agentx_parse(&session, pdu, (unsigned char *)data, size) == SNMP_ERR_NOERROR) {
+    if (agentx_parse(&session, pdu, (u_char *)((uintptr_t)data), size) ==
```
bvanassche, line 56: Please use NETSNMP_REMOVE_CONST() instead of this double cast.

bvanassche (changes requested):

testing/fuzzing/agentx_parse_fuzzer.c
DavidKorczynski: Thanks, fixed!

bvanassche:
Thanks for the patch!

PR #375, opened 2021-11-06
libsnmp: Fix stack-based buffer underflow

DavidKorczynski:
Ensure `strlen(defbuf)` is above zero before the code
`defbuf[strlen(defbuf) - 1] = 0;`
Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40699


bvanassche:
Thanks for the patch!

Issue #263, opened 2021-01-25
TLS 1.2 is not working in net-snmp 5.7.3


Nagarajan-GK:
I'm want to disable TLS 1.0 and 1.1 versions communication in net-snmp. TLS 1.2 and higher only to be allowed for communication. My system is using OpenSSL 1.0.2j.
I tried to modify net-snmp the source to support it. But, failed. 

How to do it?
Should I upgrade to latest version?

(I'm trying the same thing in net-snmp 5.9 too and not working)

hardaker:
I believe this is fixed with the previously mentioned patch.  Please re-open if this is incorrect.

PR #377, opened 2021-11-10
[TSM] TLS TCP Support the configuration of min/max tls version

lukas-wimmer:
Until now, it has been limited to TLS1.0, which is no longer considered secure. I have also increased default minimum configuration to TLS1_2

Format for configuration options (tls1 | tls1_1| tls1_2 | tls1_3) similar to openssl s_client command

### DIFF
--- a/include/net-snmp/library/default_store.h
+++ b/include/net-snmp/library/default_store.h
@@ -4,7 +4,7 @@
  */
 /*
  * Portions of this file are copyrighted by:
- * Copyright © 2003 Sun Microsystems, Inc. All rights reserved.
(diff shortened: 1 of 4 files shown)


lukas-wimmer:
Something is wrong with git here:

According changes on master  * Copyright © 2003 Sun Microsystems, Inc. All rights reserved. is deleted? But its still there?!

lukas-wimmer:
see also https://github.com/net-snmp/net-snmp/issues/263

bvanassche:
Thanks for the patch!

PR #380, opened 2021-11-18
testing/fuzzing: Add fuzzer for print functions

DavidKorczynski:
Add fuzzer which targets print_variable and snprintf functions

### DIFF
--- a/testing/fuzzing/snmp_print_var_fuzzer.c
+++ b/testing/fuzzing/snmp_print_var_fuzzer.c
@@ -1,0 +1,288 @@
+ /*
+  * Copyright (c) 2021, Net-snmp authors
+  * All rights reserved.
+  *
+  * Redistribution and use in source and binary forms, with or without
+  * modification, are permitted provided that the following conditions are met:
+  *
+  * * Redistributions of source code must retain the above copyright notice, this
+  *   list of conditions and the following disclaimer.
+  *
+  * * Redistributions in binary form must reproduce the above copyright notice,
+  *   this list of conditions and the following disclaimer in the documentation
+  *   and/or other materials provided with the distribution.
+  *
+  * * Neither the name of the copyright holder nor the names of its
+  *   contributors may be used to endorse or promote products derived from
+  *   this software without specific prior written permission.
+  *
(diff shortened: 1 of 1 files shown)


testing/fuzzing/snmp_print_var_fuzzer.c
```diff
@@ -0,0 +1,271 @@
...
+        break;
+    }
+
+    /*
+     * Fuzz snprint_* functions
+     */
+    variable.type = ASN_INTEGER;
+    long            value1 = 3;
+    long           *value1_ptr = &value1;
+    variable.val.integer = (long *) &value1_ptr;
```
bvanassche, line 251: Shouldn't this be changed into `variable.val.integer = &value1;` since the variable type is `ASN_INTEGER`?

testing/fuzzing/snmp_print_var_fuzzer.c
```diff
@@ -0,0 +1,271 @@
...
+    variable.val_len = 4;
+    variable.next_variable = NULL;
+    char            out_buf[100];
+    snprint_integer(out_buf, 100, &variable, NULL, s3, s4);
+
+    variable.type = ASN_UINTEGER;
+    long            value2 = 3;
+    long           *value2_ptr = &value2;
+    variable.val.integer = (long *) &value2_ptr;
```
bvanassche, line 260: Same comment here ...

bvanassche (changes requested):

testing/fuzzing/snmp_print_var_fuzzer.c
DavidKorczynski: Fixed! I did this essentially to accommodate for this line: https://github.com/net-snmp/net-snmp/blob/master/snmplib/mib.c#L1276 
Which formats a string based on `hints`. Since `hints` was fuzzed data, it could use a `%s` or so on the dereference `variable.val.integer`. However, now, instead I limit what `hints` can be.

testing/fuzzing/snmp_print_var_fuzzer.c
DavidKorczynski: Fixed as per above.

bvanassche:
Thanks for the patch!

PR #387, opened 2021-11-29
Add support for digests detected from ECC certificates

minfrin:
Previously, the digest could be detected on RSA certificates only. This
patch adds detection for ECC certificates.

[ bvanassche: changed _htmap2 into a two-dimensional array and renamed _htmap2
  back to _htmap ]


bvanassche:
No new releases will happen from the V5-9-1-patches branch. Is V5-9-1-patches really the branch you want to apply this patch onto? Please reopen this pull request if it needs further attention.

PR #404, opened 2022-02-10
Fix various error handling issues

bernd-edlinger:


bernd-edlinger:
Hi,

I started this work when I found lots of issues when I tried
./configure CFLAGS="-g -fsanitize=address"

... and got a lot more issues when I stated to do error injection tests:
./configure CFLAGS="-g -fsanitize=address -DERROR_INJECTION"

See the last commit 4d7f023 for how that works.
Initially I was blown away by a lot of memory leaks that
can be created by REGISTER_MIB.  I fixed a lot of issues,
but obviously still a lot remain to be done.

Thanks,
Bernd.

moshekaplan:
@bernd-edlinger , @bvanassche : Would it be helpful if I were to try to split some of the simpler changes here into many smaller PRs, so that they could be easily reviewed and merged in?

bvanassche:
Splitting this pull request into one patch per function would help. Something additional to consider is how valuable these changes are? Usually snmpd runs on systems in which much more memory is available than what snmpd needs. Additionally, every change involves the risk of introducing a bug. Applying all the changes from this pull request involves a certain risk.

moshekaplan:
Thanks @bvanassche . I think this PR has some amazing work, but the huge scope of the changes makes reviewing it daunting.

PR #417, opened 2022-04-12
don't try to install factory.h

sthen:
87d45755edd made factory.h a private header, but snmplib/Makefile.in
continues to try to install it from include/net-snmp/library, resulting
in an error from "make install".

I'm a bit surprised to run into this, so maybe I'm missing something though...


bvanassche:
Thanks for the patch!

PR #424, opened 2022-05-25
Add ignoremount config option v2

po5857:
As was attempted before (https://github.com/net-snmp/net-snmp/pull/324), I would like to add the "ignoremount" option we have been using for a few years into upstream.

We monitor many hosts, but we want the ability to ignore certain mounts for a couple of reasons:

  1. Some mounts (such as backup mounts) will always be near 100%, but we do not care.  Ignoring them in net-snmp is easier than in our monitoring system
  2. Some mounts are on systems which get so heavily used that snmpd will hang trying to check on their usage. While this is obviously a problem we should solve (fix the storage system), having snmpd ignore them is a cheaper option

The attached patch adds an "ignoremount" option to snmpd.conf which can be used like so:

ignoremount /mnt/foo
ignoremount /var/bar
ignoremount -r /var/backup/big[1-4]

This makes the snmpd agent completely skip over the configured mounts for all statistic related inquiries - they show as 0% used instead.

Changes from v1:
* Added regex and pcre support
* Moved config option parsing into generic code
* Updated manpage

Please pull, thanks.

### DIFF
--- a/agent/mibgroup/hardware/fsys/fsys_mntent.c
+++ b/agent/mibgroup/hardware/fsys/fsys_mntent.c
@@ -27,6 +27,12 @@
 #if HAVE_SYS_STATVFS_H
 #include <sys/statvfs.h>
 #endif
+#if defined(HAVE_PCRE_H)
+#include <pcre.h>
+#elif defined(HAVE_REGEX_H)
+#include <sys/types.h>
+#include <regex.h>
+#endif
 
 #ifdef solaris2
 #define _NETSNMP_GETMNTENT_TWO_ARGS 1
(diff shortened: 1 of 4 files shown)


bvanassche:
Thanks for the patch! A modified version of this patch has been applied on the master branch. Please retest.

po5857:
Tested it out and works as expected.  Thanks @bvanassche !

po5857:
@bvanassche will this be pulled into the 5.9.x branch, or is it awaiting 5.10 (or 6.0)?

bvanassche:
This patch has been cherry-picked onto the V5-9-patches branch as commit e3c1de1e3b69729a895333c9a7e9a2a99800f691.

PR #425, opened 2022-05-25
Add virtiofs to other_fs

po5857:
virtiofs is a relatively new virtual filesystem.  At present, it is misdetected by net-snmp and needs to be added to the other_fs array.


po5857:
Botched this request with ignoremount commit buried in it.

PR #430, opened 2022-06-21
libsnmp: Set Linux VRF iface on Trap sink IP addresses for TCP transport

### LINKED ISSUE #389: Set Linux VRF iface on Trap sink IP addresses for TCP transport
Hello!
Afrer patch 02de400544dea6804c348c457a5ea5ea5531b12d i can use `<proto:ip@VRF:port>` without clientaddr.
But it is only UDP.

For TCP it was missing or it was planned so?

P.S. For test I added code from patch to netsnmp_tcp_transport() and  netsnmp_tcp6_transport() and TCP  now work via VRF too  wihtout clientaddr

ak503:
It's based on the 02de400544dea6804c348c457a5ea5ea5531b12d for fixes an issue #12 with TCP transport
(fix #389)

Signed-off-by: Dmitrii Turlupov <dturlupov@factor-ts.ru>

### DIFF
--- a/snmplib/transports/snmpTCPDomain.c
+++ b/snmplib/transports/snmpTCPDomain.c
@@ -200,6 +200,17 @@ netsnmp_tcp_transport(const struct netsnmp_ep *ep, int local)
 
     t->flags = NETSNMP_TRANSPORT_FLAG_STREAM;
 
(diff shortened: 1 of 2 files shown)


bvanassche:
Thanks for the patch!

PR #439, opened 2022-07-25
testing/fuzzing: fix transport fuzzer coverage

DavidKorczynski:
This patch fixes an unexpected-exit issue
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=40640 which
is likely the cause of a coverage failure issue
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46588

The problem is netsnmp_tdomain_transport_full eventually forks the
process and also calls exit, which is not compatible with fuzzing. As
such, the function should not be called by the fuzzer.


bvanassche:
Thanks for the patch!

PR #440, opened 2022-07-26
Free new type variable. May cause memory leaks.

zxlin96:


bvanassche:
The patch description is confusing to me. Does the description mean that this patch may introduce a memory leak? If not, please rephrase the patch description.

snmplib/snmp_api.c
```diff
@@ -7140,7 +7144,7 @@ netsnmp_oid_find_prefix(const oid * in_name1, size_t len1,
     min_size = SNMP_MIN(len1, len2);
     for(i = 0; i < (int)min_size; i++) {
         if (in_name1[i] != in_name2[i])
-            return i;    /* '�' is the first differing subidentifier
+            return i;    /* '�' is the first differing subidentifier
```
fenner, line 7147: What is the goal of this change?  What is the new character inside the quotes?  Github can't render it.

snmplib/snmp_api.c
```diff
@@ -5464,6 +5464,10 @@ snmp_free_var_internals(netsnmp_variable_list * var)
         SNMP_FREE(var->name);
     if (var->val.string != var->buf)
         SNMP_FREE(var->val.string);
+    if (var->val.bitstring)
+        SNMP_FREE(var->val.bitstring);
+    if (var->val.objid)
+        SNMP_FREE(var->val.objid);
```
fenner, line 5470: Given that string, bitstring and objid are all pointers inside the same union, can you please explain how these are different than operating on `string`?

```C
typedef union {
   long           *integer;
   u_char         *string;
   oid            *objid;
   u_char         *bitstring;
   struct counter64 *counter64;
#ifdef NETSNMP_WITH_OPAQUE_SPECIAL_TYPES
   float          *floatVal;
   double         *doubleVal;
   /*
    * t_union *unionVal; 
    */
#endif                          /* NETSNMP_WITH_OPAQUE_SPECIAL_TYPES */
} netsnmp_vardata;
```

fenner (changes requested):

fenner:
Closing due to no response.  (Note that the regular tests show that nearly every test fails with this change.)

PR #446, opened 2022-08-20
Two compilation fixes for FreeBSD 14

glebius:
The first patch is actually a fix for all modern FreeBSD versions, as they compile but end up with wrong values.


bvanassche:
A slightly modified version of these patches has been applied. Thanks for the patches!

PR #448, opened 2022-08-26
Fix sast issues

ProjectMutilation:
Hi, i found some potential bugs with using PVS-Studio static analyzer. I offer the solution of these bugs


bvanassche:
Slightly modified versions of these patches have been applied on the master branch. Please review the patches that have been applied on the master branch.

ProjectMutilation:
@bvanassche 
i don't agree with patch into  commit 44d304e87cb07c36b560050f32535c7d3ec7e44c
integer overflow is undefined behavior, there is no guarantee that the value of the variable will become negative
in practice, I have seen that the value can remain int_max

bvanassche:
How about changing the type of `ipAddressSpinLockValue` to `unsigned int` and restoring the `ipAddressSpinLockValue > 2147483647` check?

ProjectMutilation:
ipAddressSpinLockValue is initialized by calling netsnmp_random that return value with long type, overflow is possible on this step. Maybe it should be long? and an overflow check should be placed before changing the value

bvanassche:
How about the patch below?
```
Subject: [PATCH] IP-MIB: Do not trigger undefined behavior

This patch reworks commit 44d304e87cb07c36b560050f32535c7d3ec7e44c.

Reported-by: Maxim Korotkov <gun.fever@yandex.ru>
---
 agent/mibgroup/ip-mib/ip_scalars.c | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/agent/mibgroup/ip-mib/ip_scalars.c b/agent/mibgroup/ip-mib/ip_scalars.c
index 1b64692d5f9e..7771838af070 100644
--- a/agent/mibgroup/ip-mib/ip_scalars.c
+++ b/agent/mibgroup/ip-mib/ip_scalars.c
@@ -36,7 +36,7 @@ handle_ipv6IpDefaultHopLimit(netsnmp_mib_handler *handler,
                              netsnmp_agent_request_info *reqinfo,
                              netsnmp_request_info *requests);
 
-static int ipAddressSpinLockValue;
+static uint32_t ipAddressSpinLockValue;
 
 static int
 handle_ipAddressSpinLock(netsnmp_mib_handler *handler,
@@ -87,7 +87,7 @@ init_ip_scalars(void)
                              HANDLER_CAN_RWRITE));
 
     /* Initialize spin lock with random value */
-    ipAddressSpinLockValue = netsnmp_random();
+    ipAddressSpinLockValue = netsnmp_random() & INT32_MAX;
 
 }
 
@@ -445,7 +445,7 @@ handle_ipAddressSpinLock(netsnmp_mib_handler *handler,
             } else {
                 ipAddressSpinLockValue++;
                 /* and check it for overflow */
-                if (ipAddressSpinLockValue < 0)
+                if (ipAddressSpinLockValue > INT32_MAX)
                     ipAddressSpinLockValue = 0;
             }
             break;
```

ProjectMutilation:
> 

``+                if (ipAddressSpinLockValue > INT32_MAX)`` is useless, int variable can't be larger than INT32_MAX
in my opinion maybe
`` if ((ipAddressSpinLockValue == INT32_MAX)||(ipAddressSpinLockValue<0))``

bvanassche:
INT32_MAX (2147483647) is the maximum for a signed integer. An unsigned integer can be larger than INT32_MAX and cannot be negative. So I think the patch in my previous comment is fine.

ProjectMutilation:
> INT32_MAX (2147483647) is the maximum for a signed integer. An unsigned integer can be larger than INT32_MAX and cannot be negative. So I think the patch in my previous comment is fine.

my bad, I didn't see that the type of the variable has changed, then the patch is ok

PR #450, opened 2022-09-05
unload_all_mibs: fix memory leak by freeing tclist

JanSoundhouse:
tclist is always allocated in netsnmp_init_mib_internals, when doing multiple init_snmp("")/snmp_shutdown("") this memory is never free'd.


bvanassche:
Thanks for the patch!

PR #452, opened 2022-09-06
feature-auth-failure-unknown-security-name

EZenderink:
Hi,

To log if authentication failures happen, not only in regards to password, we would like to add this additional feature to generate a AUTH_FAILURE callback if a unknown security name is used as well.

Please let me know if this ok.

Kind Regards,

Eldin

### DIFF
--- a/snmplib/snmp_api.c
+++ b/snmplib/snmp_api.c
@@ -7152,7 +7152,7 @@ netsnmp_oid_find_prefix(const oid * in_name1, size_t len1,
     min_size = SNMP_MIN(len1, len2);
     for(i = 0; i < (int)min_size; i++) {
         if (in_name1[i] != in_name2[i])
-            return i;    /* 'í' is the first differing subidentifier
+            return i;    /* 'i' is the first differing subidentifier
                             So the common prefix is 0..(i-1), of length i */
     }
     return min_size;	/* The shorter OID is a prefix of the longer, and
--- a/snmplib/snmp_api.c
+++ b/snmplib/snmp_api.c
@@ -111,6 +111,7 @@ SOFTWARE.
 #include <net-snmp/output_api.h>
 #include <net-snmp/config_api.h>
 #include <net-snmp/utilities.h>
+#include <net-snmp/agent/agent_callbacks.h>
 
 #include <net-snmp/library/asn1.h>
 #include <net-snmp/library/snmp.h>      /* for xdump & {build,parse}_var_op */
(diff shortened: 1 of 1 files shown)


snmplib/snmp_api.c
```diff
@@ -4378,6 +4379,11 @@ _snmp_parse(struct session_list *slp,
                     pdu->securityName, secLevelName[pdu->securityLevel],
                     snmp_api_errstring(result)));
 
+        if (result == SNMPERR_USM_UNKNOWNSECURITYNAME) {
+            snmp_call_callbacks(SNMP_CALLBACK_APPLICATION,
+                                SNMPD_CALLBACK_AUTH_FAILURE, (void *)pdu);
```
bvanassche, line 4384: Is this cast necessary?

snmplib/snmp_api.c
```diff
@@ -7146,7 +7152,7 @@ netsnmp_oid_find_prefix(const oid * in_name1, size_t len1,
     min_size = SNMP_MIN(len1, len2);
     for(i = 0; i < (int)min_size; i++) {
         if (in_name1[i] != in_name2[i])
-            return i;    /* '�' is the first differing subidentifier
```
bvanassche, line 7149: It is not clear to me what has been changed in this line?

bvanassche (changes requested):

snmplib/snmp_api.c
EZenderink: apparently the encoding of the character 'í', I can see if I can create a pull request/ change the commit to not include this change bit odd tho.. (Though I would believe the 'í' should be i actually, would that change be ok?)

snmplib/snmp_api.c
EZenderink: Sorry, no, it is not necessary (it was based on a similar callback call, but it appears the cast has been removed there as well).

snmplib/snmp_api.c
bvanassche: Thanks for the clarification. I think that change counts as a fix :-)

bvanassche:
A modified version of these patches has been applied on the master branch. Please take another look at the master branch.

EZenderink:
Hi, @bvanassche I currently do not see any new commits since the 6th of September on the master branch, or has a similar functionality been implemented in the past?

bvanassche:
Thanks for having reported this. Please take another look at the master branch.

PR #466, opened 2022-10-13
Overhaul support for the SNMPv3+SSH transport

minfrin:
Attempts to connect on RHEL9 machines would fail on sendmsg(). Avoid EINVAL when passing credentials over SSH unix domain socket.

Align the socket path with the SSH transport.

Add option to log in using an sshagent in addition to explicit keys.

Initial IPv6 support.

### DIFF
--- a/apps/sshtosnmp.c
+++ b/apps/sshtosnmp.c
@@ -32,7 +32,7 @@
 #define MAXPATHLEN 2048
 #endif
 
-#define DEFAULT_SOCK_PATH "/var/net-snmp/sshdomainsocket"
+#define DEFAULT_SOCK_PATH NETSNMP_PERSISTENT_DIRECTORY "/sshdomainsocket"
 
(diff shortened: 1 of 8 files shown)


minfrin:
Example command line:

```
/tmp/net-snmp/bin/snmpbulkwalk --defSecurityModel=tsm --defSecurityName=root --sshagent=true --sshusername=root -l authPriv -Dnone -t 10 ssh:blackadder.example.com:22
```

- You have to specify defSecurityName=root and sshusername=root. If not, you get the dreaded "failed rfc5343 contextEngineID probing" as the client side has no idea what the username is.
- Without authPriv you get "Reason: authorizationError (access denied to that object)".
- Without --defSecurityModel=tsm you get "netsnmp_ssh_send wasn't passed a valid tmStateReference"

Obvs work outstanding to fix these glitches.

bvanassche:
A slightly modified version of these patches has been applied on the master branch. Thank you for your patience!

PR #482, opened 2022-11-21
Snmpd: fix potential null pointer dereference

ProjectMutilation:


agent/snmpd.c
```diff
@@ -494,10 +494,10 @@ main(int argc, char *argv[])
     while ((arg = getopt(argc, argv, options)) != EOF) {
         switch (arg) {
         case '-':
-            if (strcasecmp(optarg, "help") == 0) {
+            if (optarg && (strcasecmp(optarg, "help") == 0)) {
```
bvanassche, line 497: `optarg` cannot be NULL here since `-` in the option string is followed by `:`.

bvanassche:
I think the current code is fine and hence that the proposed checks are not needed. So I will close this pull request.

PR #484, opened 2022-11-23
Snmptrapd cannot reconnect to MySQL server after hitting MySQL wait_timeout

yukariatlas:
**[Environment]**
MySQL server v8.0.31
Snmptrapd v5.10

**[Issue]**
wait_timeout is the number of seconds the MySQL server waits for activity before closing the connection.
Snmptrapd cannot reconnect to the MySQL server once the connection is closed.

**[Root Cause]**
Snmptrapd can reconnect to the MySQL server by using MySQL v5.7.40.
The new version of MySQL writes the reason for the connection before closing it, and the client receives a more informative error message (ER_CLIENT_INTERACTION_TIMEOUT).
Snmptrapd does not handle this error code, so the connection will not reconnect to the MySQL server.

**[Solution]**
The implementation of mysql_real_connect may be changed, so only handling the error code can not solve the problem.
We should close the existing connection and reinitialize it in netsnmp_mysql_connect function.
The default value of wait_timeout is 8 hours, which may not increase too much performance overhead by this approach.

Any comments and suggestions are welcome!

### DIFF
--- a/apps/snmptrapd_sql.c
+++ b/apps/snmptrapd_sql.c
@@ -256,6 +256,13 @@ netsnmp_sql_disconnected(void)
     }
 }
 
+static bool
+netsnmp_sql_server_disconnected(int err)
+{
+    // CR_SERVER_GONE_ERROR | CR_SERVER_LOST | ER_CLIENT_INTERACTION_TIMEOUT
+    return CR_SERVER_GONE_ERROR == err || CR_SERVER_LOST == err || 4031 == err;
+}
+
 /*
  * convenience function to log mysql errors
  */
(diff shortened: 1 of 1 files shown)


yukariatlas:
The patch handles three error codes (i.e., CR_SERVER_GONE_ERROR/CR_SERVER_LOST/ER_CLIENT_INTERACTION_TIMEOUT) needed to be reconnected.
It also resents sql requests if the previous failure is a connection timeout.
The issue will be significant if users modify the MySQL wait_timeout or snmp events are triggered infrequently.
@bvanassche Could you please spend some time to take a look and give me some feedback if there is anything improper?

Any comments and suggestions are welcome!

bvanassche:
Thanks for the patch!

PR #485, opened 2022-11-23
mibgroup: fix possible out-of-bound read

ProjectMutilation:


fenner (commented):
Can you please share the scenario in which cptr is NULL and i == 0, which is what you are trying to protect against?

It's called indirectly from
```
            if (cptr == NULL) {
                netsnmp_config_error("Blank line following %s token.", token);
            } else {
                DEBUGMSGTL(("read_config:line", "%s:%d examining: %s\n",
                            filename, linecount, line));
                run_config_handler(lptr, token, cptr, when);
            }
```
so - when can `run_config_handler()` be called with `cptr == NULL`?

Even if `run_config_handler()` is called via a different path with `cptr == NULL`, it has this code before calling the next layer handler such as `loadave_parse_config()`:
```
            /*
             * Make sure cptr is non-null
             */
            if (!cptr) {
                tmpbuf[0] = '\0';
                cptr = tmpbuf;
            }
```

This pull request is trying to address a situation that does not happen: that cptr is passed in as NULL.

`skip_not_white()` or `skip_white()` can return NULL, so, cptr can be NULL when i is 1 or 2, which the current code handles.

ProjectMutilation:
agree that this scenario is unlikely, but I would not rule out the possibility of its occurrence over time
But if you do not agree with the patch, I propose to close the PR

PR #486, opened 2022-11-24
snmptrapd: fix potential NULL pointer dereference

ProjectMutilation:
Pointer 'handler' which was dereferenced at snmptrapd_handlers.c:811 is compared to NULL value at snmptrapd_handlers.c:812.

Found by RASU JSC with SVACE static analyzer
Signed-off-by: Maxim Korotkov <korotkov.maxim.s@gmail.com>

### DIFF
--- a/apps/snmptrapd_handlers.c
+++ b/apps/snmptrapd_handlers.c
@@ -810,9 +810,10 @@ int   command_handler( netsnmp_pdu           *pdu,
     size_t          r_len = 64, o_len = 0;
     int             oldquick;
 
+
+    if (handler && handler->token && *handler->token) {
     DEBUGMSGTL(( "snmptrapd", "command_handler\n"));
     DEBUGMSGTL(( "snmptrapd", "token = '%s'\n", handler->token));
-    if (handler && handler->token && *handler->token) {
 	netsnmp_pdu    *v2_pdu = NULL;
 	if (pdu->command == SNMP_MSG_TRAP)
 	    v2_pdu = convert_v1pdu_to_v2(pdu);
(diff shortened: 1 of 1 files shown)


apps/snmptrapd_handlers.c
```diff
@@ -810,9 +810,10 @@ int   command_handler( netsnmp_pdu           *pdu,
     size_t          r_len = 64, o_len = 0;
     int             oldquick;
 
+
+    if (handler && handler->token && *handler->token) {
```
bvanassche, line 814: Shouldn't this if-statement be swapped with the debug statement below it? Please also fix the indentation.

bvanassche (changes requested):

apps/snmptrapd_handlers.c
ProjectMutilation: Debug has been replaced for avoiding dereference of handler->token that can be NULL

fenner: handler can not be NULL.  The calling sequence is

```
                ret = (*(traph->handler))(pdu, transport, traph);
```

The `handler` arg is the `traph` variable here.  If `traph` is NULL, this code will fail without calling the handler.  Therefore, the right fix is to instead remove the `handler &&` check, to avoid the appearance that it's possible for handler to be NULL.

bvanassche: The above suggestion has been implemented in commit dc4ef4a5aed66f2e2048657c2c95413b3f294035. Does everyone agree that that commit is sufficient to close this pull request?

PR #487, opened 2022-11-24
agent: fix potential null ptr dereference

ProjectMutilation:
The pointer vp was dereferenced before it was checked againt NULL at case MODE_GETNEXT

Signed-off-by: Maxim Korotkov <korotkov.maxim.s@gmail.com>

### DIFF
--- a/agent/helpers/old_api.c
+++ b/agent/helpers/old_api.c
@@ -278,7 +278,7 @@ netsnmp_old_api_helper(netsnmp_mib_handler *handler,
     int             exact = 1;
     int             status;
 
-    struct variable *vp;
+    struct variable *vp = NULL;
     netsnmp_old_api_cache *cacheptr;
     netsnmp_agent_session *oldasp = NULL;
     u_char         *access = NULL;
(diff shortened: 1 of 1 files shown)


agent/helpers/old_api.c
```diff
@@ -278,20 +278,23 @@ netsnmp_old_api_helper(netsnmp_mib_handler *handler,
...
     netsnmp_old_api_cache *cacheptr;
     netsnmp_agent_session *oldasp = NULL;
     u_char         *access = NULL;
     WriteMethod    *write_method = NULL;
     size_t          len;
     size_t          tmp_len;
     oid             tmp_name[MAX_OID_LEN];
-
-    vp = (struct variable *) handler->myvoid;
+    if(handler)
```
bvanassche, line 289: Hmm ... how could handler be NULL?

bvanassche (changes requested):

agent/helpers/old_api.c
ProjectMutilation: it's was added for be on the safe side

fenner: > it's was added for be on the safe side

By that philosophy, every access of every pointer anywhere should be NULL-checked.

I agree with Bart, that handler can not be null - the API defines this.

I'm also curious by which path handler->myvoid can be NULL, because it is unconditionally set in `netsnmp_register_old_api()` before calling `netsnmp_register_handler()`.  @ProjectMutilation , can you explain the code path that results in handler->myvoid being NULL?  Do you have a series of API calls or a packet that can cause this?

ProjectMutilation: I assumed that the vp pointer is checked for NULL lower in the code(328 line at original) and therefore can be NULL.  But vp was dereferenced before cheching( 301 line at original)

hardaker: (on the other hand -- it doesn't really hurt to include other error checking)

PR #488, opened 2022-11-24
mibgroup: fix logic error

ProjectMutilation:
linux_read_mibII_stats() returned 0 if
ret is successful, or -1 if ret is fail

### DIFF
--- a/agent/mibgroup/mibII/kernel_linux.c
+++ b/agent/mibgroup/mibII/kernel_linux.c
@@ -438,9 +438,6 @@ linux_read_icmp_msg_stat(struct icmp_mib *icmpstat,
 
     if ((ret = linux_read_mibII_stats()) == -1) {
         return -1;
-    } else if (ret) {
-       memcpy(icmpmsgstat, &cached_icmp4_msg_mib, sizeof(*icmpmsgstat));
-       *flag = 1; /* We have a valid icmpmsg */
     }
 
     memcpy(icmpstat, &cached_icmp_mib, sizeof(*icmpstat));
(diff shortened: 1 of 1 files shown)


agent/mibgroup/mibII/kernel_linux.c
```diff
@@ -438,7 +438,7 @@ linux_read_icmp_msg_stat(struct icmp_mib *icmpstat,
 
     if ((ret = linux_read_mibII_stats()) == -1) {
         return -1;
-    } else if (ret) {
+    } else if (!ret) {
```
bvanassche, line 441: Please leave out the "else if (ret)" instead of modifying it.

ProjectMutilation: it's dead code, the condition is always false,  is it works as designed?

fenner: Correct, the case where ret=1 was returned was removed in 142c4239771c6282d66a553d5692037c251a2f90

ProjectMutilation: > Correct, the case where ret=1 was returned was removed in [142c423](https://github.com/net-snmp/net-snmp/commit/142c4239771c6282d66a553d5692037c251a2f90)

maybe then delete this case?

fenner: Correct, that is what Bart meant by "leave it out".

ProjectMutilation: sorry, apparently there was a misunderstanding, I thought just leave without changes

bvanassche:
A modified version of patch "mibgroup: removed dead code" has been applied. Please take a look.

PR #491, opened 2022-11-28
apps: fix handle leak

ProjectMutilation:
In default case dest wasn't freed, although in the current code this is hardly possible, here is potential problem


bvanassche:
Thanks for the patch!

PR #493, opened 2022-12-02
Fix LDFLAGS vs LIBS ordering

thesamesam:
LDFLAGS must come before both LIBS & any listed objects in order
for certain valid LDFLAGS, like '-Wl,--as-needed' to work correctly
(otherwise it'll either take no effect or discard libraries when
they're needed).

Gentoo has been carrying this patch for a while.

Signed-off-by: Sam James <sam@gentoo.org>


bvanassche:
A slightly modified version of this patch series has been applied. See also commit ae9eb483e607a035680d5429728ef637139690bc and before.

thesamesam:
Thank you again!

PR #494, opened 2022-12-02
configure: Fix false positive when build dir contains "yes"

inguin:
The checks for various IPv6 libraries succeed when the word "yes" appears anywhere in the CPP output. That's prone to false positives, e.g. when the build environment path contains that string.

Make the check more robust by anchoring the pattern so the word must appear on a line of its own.


bvanassche:
A slightly modified version of this patch has been applied. Please take a look at commit a5f46cefffdf5e0ca011aff5f6414f74eed8e909.

PR #495, opened 2022-12-09
apps: combining conditions

ProjectMutilation:
I think it's better to remove the extra branching in the condition and merge them


bvanassche:
Thanks for the patch!

PR #497, opened 2022-12-13
[5.9.3]: Ipaddress linux unbreak without rtnetlink

jacmet:
Commit 89a7860829fdc618ef (IP-MIB: Fix the --disable-ipv6 build) moved defines around to fix builds without IPv6 support, but accidently caused netsnmp_access_other_info_get() to be visible even if HAVE_LINUX_RTNETLINK_H isn't defined, leading to build issues:

mibgroup/ip-mib/data_access/ipaddress_linux.c: In function ‘netsnmp_access_other_info_get’: mibgroup/ip-mib/data_access/ipaddress_linux.c:439:28: error: field ‘n’ has incomplete type
            struct nlmsghdr n;
                            ^
mibgroup/ip-mib/data_access/ipaddress_linux.c:440:29: error: field ‘r’ has incomplete type
            struct ifaddrmsg r;
                             ^
mibgroup/ip-mib/data_access/ipaddress_linux.c:454:40: error: ‘NETLINK_ROUTE’ undeclared (first use in this function)
    sd = socket(PF_NETLINK, SOCK_DGRAM, NETLINK_ROUTE);
                                        ^~~~~~~~~~~~~
mibgroup/ip-mib/data_access/ipaddress_linux.c:454:40: note: each undeclared identifier is reported only once for each function it appears in mibgroup/ip-mib/data_access/ipaddress_linux.c:461:22: warning: implicit declaration of function ‘NLMSG_LENGTH’; did you mean ‘CMSG_LEN’? [-Wimplicit-function-declaration]
    req.n.nlmsg_len = NLMSG_LENGTH(sizeof(struct ifaddrmsg));
                      ^~~~~~~~~~~~
                      CMSG_LEN
mibgroup/ip-mib/data_access/ipaddress_linux.c:461:42: error: invalid application of ‘sizeof’ to incomplete type ‘struct ifaddrmsg’
    req.n.nlmsg_len = NLMSG_LENGTH(sizeof(struct ifaddrmsg));
                                          ^~~~~~

Fix it by also enclosing it within the RTNETLINK and IPV6 guards, similar to netsnmp_access_ipaddress_extra_prefix_info().  Given that are already inside an IPV6 conditional just above and below, simply reuse that for simplicity.

Signed-off-by: Peter Korsgaard <peter@korsgaard.com>

### DIFF
--- a/agent/mibgroup/ip-mib/data_access/ipaddress_ioctl.c
+++ b/agent/mibgroup/ip-mib/data_access/ipaddress_ioctl.c
@@ -244,7 +244,7 @@ _netsnmp_ioctl_ipaddress_container_load_v4(netsnmp_container *container,
          * get broadcast
          */
         memset(&addr_info, 0, sizeof(struct address_flag_info));
-#if defined (NETSNMP_ENABLE_IPV6)
+#if defined (NETSNMP_ENABLE_IPV6) && defined(HAVE_LINUX_RTNETLINK_H)
         addr_info = netsnmp_access_other_info_get(entry->if_index, AF_INET);
         if(addr_info.bcastflg) {
            bcastentry = netsnmp_access_ipaddress_entry_create();
--- a/agent/mibgroup/ip-mib/data_access/ipaddress_linux.c
+++ b/agent/mibgroup/ip-mib/data_access/ipaddress_linux.c
@@ -430,8 +430,8 @@ _load_v6(netsnmp_container *container, int idx_offset)
     return idx_offset;
 #endif /* HAVE_LINUX_RTNETLINK_H */
 }
-#endif /* NETSNMP_ENABLE_IPV6 */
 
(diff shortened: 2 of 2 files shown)


ThomasDevoogdt:
The only problem here is that it also needs to be fixed in ipaddress_ioctl, so we should add a HAVE_LINUX_RTNETLINK_H there as well.
https://github.com/net-snmp/net-snmp/blob/0608dce6c4f4b5a978672e958047d1098137e2a7/agent/mibgroup/ip-mib/data_access/ipaddress_ioctl.c#L247-L261

jacmet:
@ThomasDevoogdt ups indeed, I got confused by a local patch already doing that. Fixed now - Thanks!

bvanassche:
These patches have been applied on the v5.9 and master branches. Thanks for the patches!

PR #498, opened 2022-12-14
snmplib: fix memory leak

ProjectMutilation:
Fixed the incorrect error handling that was the cause of memory leak

Found by RASU JSC with ISP RAS Svace
Signed-off-by: Maxim Korotkov <maskorotkov@rasu.ru>

### DIFF
--- a/snmplib/parse.c
+++ b/snmplib/parse.c
@@ -2215,8 +2215,12 @@ parse_enumlist(FILE * fp, struct enum_list **retp)
     *retp = ep;
     return ep;
     out_err:
-        free(epp);
-        return NULL;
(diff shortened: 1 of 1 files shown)


bvanassche:
I think there are three issues with this pull request:
* `free(epp);` should be changed into `free(*epp);`
* `*epp = NULL` should be added after the `free(*epp)` call.
* `if (*epp) free((*epp)->label);' should be added above `free(*epp);`

bvanassche:
A modified version of this pull request has been applied as commit ee5fca70b03115fef5f0207fc6aa6fd42b7302ea. Please take a look at that commit.

PR #501, opened 2022-12-22
mibgroup: fix memleak in netsnmp_access_interface_container_init()

ProjectMutilation:
Fixed incorrect error handling that was the cause of memory leak 
Found by RASU JSC with SVACE


bvanassche:
A modified version of this patch has been applied as commit db4136680ca9375ead66dbdaf58a8955bb12bf2e. Please take a look.

PR #502, opened 2022-12-23
mibgroup: fix NULL dereference in write_snmpNotifyFilterProfileRowStatus()

ProjectMutilation:
In case RESERVE_1 pointer var_val was checked against NULL, although it was dereferenced for assignment set_value before check
Found by RASU JSC

### DIFF
--- a/agent/mibgroup/notification/snmpNotifyFilterProfileTable.c
+++ b/agent/mibgroup/notification/snmpNotifyFilterProfileTable.c
@@ -395,6 +395,7 @@ write_snmpNotifyFilterProfileRowStatus(int action,
     static struct snmpNotifyFilterProfileTable_data *StorageDel;
     size_t          newlen = name_len - table_offset;
     static int      old_value;
+    if (!var_val) return SNMP_ERR_WRONGTYPE; 
     int             set_value = *((long *) var_val);
     netsnmp_variable_list *vars;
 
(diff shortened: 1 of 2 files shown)


hardaker:
As is, this writes a code block into the middle of variable declarations.  Although some newer C compilers allow for this, we still target ones that may not -- thus the code must go below all variable definitions.

ProjectMutilation:
@hardaker okay, If you agree with the need for a patch and there are no other comments, I will correct for compliance.

Thanks for the feedback

bvanassche:
I'm not sure this patch is the right approach to suppress the static analyzer warning. Shouldn't the caller guarantee that var_val != NULL? If so, a better fix would be to remove the var_val != NULL checks instead of adding another such check.

bvanassche:
> > I'm not sure this patch is the right approach to suppress the static analyzer warning. Shouldn't the caller guarantee that var_val != NULL? If so, a better fix would be to remove the var_val != NULL checks instead of adding another such check.
> 
> I agree that it's probably just an unnecessary check, but is it for sure?

My conclusion from a review from the code in the snmplib/ directory is that either var_val != NULL && var_len != 0 or that var_val == NULL && var_len == 0.

ProjectMutilation:
> > > I'm not sure this patch is the right approach to suppress the static analyzer warning. Shouldn't the caller guarantee that var_val != NULL? If so, a better fix would be to remove the var_val != NULL checks instead of adding another such check.
> > 
> > 
> > I agree that it's probably just an unnecessary check, but is it for sure?
> 
> My conclusion from a review from the code in the snmplib/ directory is that either var_val != NULL && var_len != 0 or that var_val == NULL && var_len == 0.

okay, then PR is unnecessary and can be closed

PR #507, opened 2023-01-05
Fix use of AC_EGREP_CPP or avoid it entirely.

iazz:
Using AC_EGREP_CPP is prone to false positives if not used carefully.

Use AC_CHECK_TYPE when looking for struct ortentry instead.

Otherwise, quote the first argument to AC_EGREP_CPP, for more readability and consistency.

### DIFF
--- a/configure
+++ b/configure
@@ -16374,7 +16374,7 @@ yes
     ]
 _ACEOF
 if (eval "$ac_cpp conftest.$ac_ext") 2>&5 |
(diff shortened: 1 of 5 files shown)


bvanassche:
A slightly modified version of this patch has been applied on the master branch (commit 5ee9d887bc59d455e7c03fabd306fdbe1b040680). Please take a look.

PR #511, opened 2023-01-09
Added option to exclude *md* devices in diskio.

Evil2000:
In many cases, on systems which have md raid, net-snmp reports each single disk plus the /dev/md* device in the diskIOTable. Sometimes this is not needed. This code change adds an option to exclude md entries in the diskIOTable.


Evil2000:
How about this pull request? Is someone of the staff looking at it? Could someone please approve the workflow?

hardaker:
thanks and sorry for the delay.

PR #519, opened 2023-01-19
Use the typed wrappers for STACK_OF(X509)

davidben:
This avoids the need for all the clumsy (and unchecked) pointer casts. OpenSSL also since deprecated the old names for the symbols (the underlying sk_num is now OPENSSL_sk_num), so this removes references to some deprecated names.


bvanassche:
Thanks for the patch!

PR #521, opened 2023-01-23
Restore using libtool's -no-undefined flag on Cygwin

jon-turney:
Commit cb9b87d9b6e2 ("Always use -Wl,-no-undefined when linking a shared library") swaps the libtool flag -no-undefined for the linker flag -Wl,-no-undefined on Cygwin.

These two aren't the same: ld's -no-undefined flag [1] makes unresolved references an error [1], whereas libtool's -no-undefined flag [2] assures it that there are no undefined references and therefore it may try to build a shared library on platforms where that is required to build a shared library.

[1] https://sourceware.org/binutils/docs-2.40/ld/Options.html
[2] https://www.gnu.org/software/libtool/manual/html_node/Link-mode.html#Link-mode

This flag possibly should also be used for MinGW builds, if shared libraries are wanted there.

### DIFF
--- a/configure
+++ b/configure
@@ -16302,11 +16302,16 @@ esac
 
 
 #
-#    Whether the linker supports the flag -Wl,-no-undefined. Do not use
-#    -Wl,-no-undefined on OpenBSD because it breaks linking of shared
-#    libraries. Use -Wl,-no-undefined on all other platforms such that
-#    undefined symbols are detected at compile time instead of at runtime.
+#    Whether the linker supports the flag -Wl,-no-undefined.
 #
+#    Do not use -Wl,-no-undefined on OpenBSD because it breaks linking of shared
+#    libraries. Use -Wl,-no-undefined on all other platforms such that undefined
+#    symbols are detected at compile time instead of at runtime.
+#
+#    On Windows targets, due to limitations of the PE executable format, the
+#    linker will always operate as if -Wl,-no-undefined was supplied.  Also
+#    supply the libtool flag -no-undefined to authorize it to build shared
+#    libraries.
 
 case x$target_os in
   xopenbsd*)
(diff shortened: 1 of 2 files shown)


bvanassche:
Thanks for the feedback. How about replacing the patch in this pull request with the patch below? Does the patch below work on Cygwin?
```
diff --git a/configure b/configure
index 40ecf10bda4d..242ad91e3533 100755
--- a/configure
+++ b/configure
@@ -16338,12 +16338,13 @@ fi
 rm -f core conftest.err conftest.$ac_objext \
     conftest$ac_exeext conftest.$ac_ext
       LDFLAGS="$saved_LDFLAGS"
-
+      LD_NO_UNDEFINED="-no-undefined $LD_NO_UNDEFINED"
     fi
     ;;
 esac
 
 
+
 #
 #   Socket library: either 'default' (non-MinGW) or 'winsock2' (MinGW).
 #
diff --git a/configure.d/config_os_progs b/configure.d/config_os_progs
index d1aaaa4d4240..22f9fc379b34 100644
--- a/configure.d/config_os_progs
+++ b/configure.d/config_os_progs
@@ -205,10 +205,11 @@ case x$target_os in
 		      LD_NO_UNDEFINED=-Wl,-no-undefined],
 		     [AC_MSG_RESULT([no])])
       LDFLAGS="$saved_LDFLAGS"
-      AC_SUBST(LD_NO_UNDEFINED)
+      LD_NO_UNDEFINED="-no-undefined $LD_NO_UNDEFINED"
     fi
     ;;
 esac
+AC_SUBST(LD_NO_UNDEFINED)
 
 
 #
```

bvanassche:
Thanks for the patch!

jon-turney:
> How about replacing the patch in this pull request with the patch below? Does the patch below work on Cygwin?

Yes, that also works.

PR #532, opened 2023-02-23
Cygwin uses ':' for the path separator

jon-turney:
Cygwin uses a UNIX-style ':' separator  for paths in environment variables.

This is a patch (originally by David Rothenberger), which we've been shipping in the Cygwin net-snmp package since at least 2013.


bvanassche:
Thanks for the patch! A slightly modified patch has been applied. Please take a look at commit 91673b10bac84230d8e4964abfab39d31a4e736f.

jon-turney:
Thanks. 
Apologies for forgetting to regenerate ./configure (our packaging system does this automatically after the patch is applied)

PR #534, opened 2023-02-24
libsnmp: Export netsnmp_set_mib_directory so it can be used inside a custom application.

### LINKED ISSUE #530: Question about netsnmp_set_mib_directory
Hi,

I wanted to add a MIB directory per code in my own application, but was unable to do so, since `netsnmp_set_mib_directory` is not dll-exported on windows build.

My question is:
- Is there a specific reason why `netsnmp_set_mib_directory` is not exported, but `netsnmp_get_mib_directory` is?
- Should it be exported?
- There is a `add_mibdir` which is exported, but I didnt get that to work to find my mibs. Neither before/after calling `init_snmp`

Flow86:
netsnmp_get_mib_directory was already exported, but netsnmp_set_mib_directory was not.

Now any external application can set its MIB dirs without reimplementing the netsnmp_set_mib_directory function.

closes #530

### DIFF
--- a/include/net-snmp/library/mib.h
+++ b/include/net-snmp/library/mib.h
@@ -120,6 +120,7 @@ SOFTWARE.
     NETSNMP_IMPORT
     void            print_ascii_dump(FILE *);
     void            register_mib_handlers(void);
+    NETSNMP_IMPORT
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #540, opened 2023-03-09
add '@' netgroup functionality

abergmann:
The openSUSE/SUSE net-snmp v5.7 package had a feature to use netgroups defined in `/etc/netgroup` or NIS/LDAP to gain access to a SNMP host. This feature was ported to our v5.9 and I would like to get it into the **master** branch.

How this feature works:

```
/etc/netgroup
nodes (node1,,) (node2,,) (node3,,)

/etc/snmp/snmpd.conf
rocommunity public @nodes
```

This makes it easier to define host groups that should have access to the SNMP host.


configure.d/config_os_functions
```diff
@@ -25,15 +25,16 @@ AC_TYPE_SIGNAL
...
-AC_CHECK_FUNCS([asprintf        closedir        fgetc_unlocked   ] dnl
-               [flockfile       funlockfile     getipnodebyname  ] dnl
-               [gettimeofday    getlogin                         ] dnl
-               [if_nametoindex  mkstemp                          ] dnl
-               [opendir         readdir         regcomp          ] dnl
-               [setenv          setitimer       setlocale        ] dnl
-               [setsid          snprintf        strcasestr       ] dnl
-               [strdup          strerror        strncasecmp      ] dnl
-               [sysconf         times           vsnprintf        ] )
+AC_CHECK_FUNCS([asprintf         closedir         endnetgrent  ] dnl
```
bvanassche, line 28: Please rework this change such that the number of changed lines is reduced.

snmplib/transports/snmpUDPDomain.c
```diff
@@ -98,6 +98,58 @@ netsnmp_udp_fmtaddr(netsnmp_transport *t, const void *data, int len)
     return netsnmp_ipv4_fmtaddr("UDP", t, data, len);
 }
 
+static int
```
bvanassche, line 101: Please split this change into two patches: one patch that introduces the netsnmp_udp_resolve_source() function and a second patch that adds the '@' netgroup functionality.

bvanassche (changes requested):

configure.d/config_os_functions
abergmann: The order inside the library `AC_CHECK_FUNCS` is alphabetical. So either we have to change the amount to lines as above or we do not order the functions alphabetical and add one line like this:
```
+    [endnetgrent   getnetgrent  setnetgrent    ] dnl
```

bvanassche: Please preserve the alphabetical order.

I think that adding a line with fewer than three function names is fine.

abergmann:
I accidentally closed this PR by doing a hard reset on my feature branch. Opening a new PR now.

PR #541, opened 2023-03-09
Scan the code base for "#if HAVE_XXX" and similar contructs, and fix …

nielsb:
…them

to use #ifdef

This fixes a bunch of -Wundef warnings.

### DIFF
--- a/snmplib/snmp-tc.c
+++ b/snmplib/snmp-tc.c
@@ -7,18 +7,18 @@
 #include <net-snmp/net-snmp-config.h>
 #include <net-snmp/net-snmp-features.h>
 #include <sys/types.h>
-#if HAVE_STDLIB_H
+#ifdef HAVE_STDLIB_H
 #include <stdlib.h>
 #endif
 #include <ctype.h>
-#if HAVE_STRING_H
+#ifdef HAVE_STRING_H
 #include <string.h>
 #endif
 #ifdef HAVE_NETINET_IN_H
 #include <netinet/in.h>
 #endif
 
-#if TIME_WITH_SYS_TIME
+#ifdef TIME_WITH_SYS_TIME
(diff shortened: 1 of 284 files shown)


snmplib/snmp-tc.c
```diff
@@ -255,7 +255,7 @@ ctime_to_timet(const char *str)
      */
 
 #ifdef HAVE_STRUCT_TM_TM_ISDST
-#if HAVE_DECL_DAYLIGHT
```
bvanassche, line 258: This change seems incorrect to me. I'm concerned this change will cause a compiler error if HAVE_DECL_DAYLIGHT is not defined.

snmplib/snmp-tc.c
```diff
@@ -255,7 +255,7 @@ ctime_to_timet(const char *str)
...
+#if HAVE_DECL_DAYLIGHT==1
```
bvanassche, line 258: This change seems incorrect to me. I'm concerned this change will cause a compiler error if HAVE_DECL_DAYLIGHT is not defined.

bvanassche (changes requested):

snmplib/snmp-tc.c
nielsb: It seems that the AC_CHECK_DECL will always #define this as either 0 or 1. I originally had changed it into a #ifdef but that caused errors. It has compiled on various Linux and *BSD systems like this.

bvanassche: If HAVE_DECL_DAYLIGHT is always 0 or 1, does this mean that the above change is not necessary?

nielsb: In principle, yes. I did it as a kind of warning to my self - and others - to not change that #if into an #ifdef, because that will fail.

PR #543, opened 2023-03-13
Set flags in MakefileSubs.pm at configure time

cmadamsgit:
Setting the flags at configure time instead of runtime allows use of the perl module without pulling in net-snmp-config (which in the Fedora/RHEL RPM setup also pulls in gcc and *-devel packages, not otherwise needed when just using the perl SNMP module).  The in-tree module is updated with the rest of the net-snmp code anyway, so setting this at configure time should stay consistent.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -140,6 +140,7 @@ AC_CONFIG_FILES([net-snmp-create-v3-user:net-snmp-create-v3-user.in],
 AC_CONFIG_FILES([netsnmp.pc:netsnmp.pc.in netsnmp-agent.pc:netsnmp-agent.pc.in])
 AC_CONFIG_FILES([dist/generation-scripts/gen-variables:dist/generation-scripts/gen-variables.in])
 AC_CONFIG_FILES([local/snmpconf])
+AC_CONFIG_FILES([perl/MakefileSubs.pm])
 
 AC_CONFIG_COMMANDS([default], echo timestamp > stamp-h)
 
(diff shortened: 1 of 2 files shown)


hardaker:
Interesting idea, and it should be sufficient when building within the net-snmp source tree.  The problem is that the code is also designed to be built from outside the source-tree when distributed from CPAN, etc.  I suspect many packages in distributions build it this way (for example).

cmadamsgit:
Hmm, I didn't realize that. Looks like the CPAN distribution hasn't been updated in quite some time. I know Fedora/RHEL/etc. all build the perl module from the net-snmp tree.

Even in the standalone distribution case, it'd probably be better to get the config at build time (it's linked against the library, so changing the library build options without rebuilding the perl module seems like an unusual case). Is there a good way to do that with the Makefile.PL?

cmadamsgit:
So looking at this closer... it looks like MakefileSubs.pm is only used at build time, by the various Makefile.PL files. Maybe it shouldn't be installed at all? That would eliminate the problem.

hardaker:
Well, I do actually now remember that we're no longer publishing to CPAN and only supposedly letting people install from the net-snmp source tree because anything else leaves to versioning problems.

cmadamsgit:
Huh, it worked for me. But like I said, it's not actually the right approach - the right approach is to stop installing MakefileSubs.pm at all. It's purely an artifact used to configure the build of the perl module and never should have been installed to begin with.

I'm just not sure how to do that with ExtUtils::MakeMaker.

PR #544, opened 2023-03-20
Fix selection of diskio implementation for NetBSD/OpenBSD

nielsb:


bvanassche (changes requested):
Can you please fix the typo in the patch description ("selectyion")?

PR #547, opened 2023-03-27
Added aarch64 as a recognizable CPU type

espenkn:
I don't know enough about net-snmp to tell if this is enough to make Arm64 CPU family work correctly but it will at least let me build the software.


bvanassche:
@hardaker Recent versions of autoconf support the aarch64 architecture. How about upgrading both the v5.9 and master branches from autoconf 2.69 to autoconf 2.71 (the latest version)?

PR #559, opened 2023-04-21
V5 9 patches

nielsb:
A couple of BSD related fixes, and some backports from master to allow testing with -fsanitize=address


bvanassche:
These patches have been applied (in a slightly different order).

PR #570, opened 2023-05-22
-Add NETSNMP_OID_OUTPUT_FULL_AND_NUMERIC for -OF formatting of OIDs

rgumpertz68:


rgumpertz68:
It appears that some whitespace cleanup crept in to mib.c.  I hope this won't be a problem.

PR #576, opened 2023-05-24
Fix #564 - catch for EINVAL and ENETUNREACH based on kernel version

### LINKED ISSUE #564: Kernel change from -EINVAL to -ENETUNREACH
Hi,

when sending an SNMP request to a broadcast address like 10.0.0.255 (in a /24 network), snmpd returns nothing and outputs following:   snmpd[xxxx]: send response: Failure in sendto

From further investigation it seems kernel changed its behavior ([Kernel patch](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=595e0651d029)) so sendmsg() now returns ENETUNREACH instead of EINVAL for broadcast addresses.

Would you agree, that simple switch from EINVAL to ENETUNREACH in affected file below would resolve the issue?

https://github.com/net-snmp/net-snmp/blob/4b4527a4a53a66c8e62ec5e9ae156c9a9f8e8606/snmplib/transports/snmpUDPBaseDomain.c#L318

What are your thoughts on this?

jridky:
Proposed patch should fix issue described in #564 taking different kernel versions into consideration.

### DIFF
--- a/snmplib/transports/snmpUDPBaseDomain.c
+++ b/snmplib/transports/snmpUDPBaseDomain.c
@@ -315,7 +315,7 @@ int netsnmp_udpbase_sendto_unix(int fd, const struct in_addr *srcip,
                         sizeof(struct sockaddr));
         else
             rc = sendmsg(fd, &m, MSG_DONTWAIT);
-        if (rc >= 0 || errno != EINVAL)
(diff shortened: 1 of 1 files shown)


bvanassche:
Please add a reference to the kernel commit that changed the errno value in the commit description.

PR #580, opened 2023-05-28
- Keep Windows MSVC library from generating ugly strings (it treats c…

rgumpertz68:
…haracters above 127 as isprint)

### DIFF
--- a/snmplib/mib.c
+++ b/snmplib/mib.c
@@ -647,9 +647,15 @@ sprint_realloc_octet_string(u_char ** buf, size_t * buf_len,
     case NETSNMP_STRING_OUTPUT_GUESS:
         hex = 0;
         for (cp = var->val.string, x = 0; x < (int) var->val_len; x++, cp++) {
+#if defined(_MSC_VER) /* Windows MSVC library includes non-ascii characters (above 0x7F) as printing, which can generate ugly strings */
+            if ((!isprint(*cp) || !isascii(*cp)) && !isspace(*cp) ) {
+                hex = 1;
+            }
+#else /* defined(_MSC_VER) */
             if (!isprint(*cp) && !isspace(*cp)) {
                 hex = 1;
             }
+#endif /* defined(_MSC_VER) */
         }
         break;
 


snmplib/mib.c
```diff
@@ -647,9 +647,15 @@ sprint_realloc_octet_string(u_char ** buf, size_t * buf_len,
     case NETSNMP_STRING_OUTPUT_GUESS:
         hex = 0;
         for (cp = var->val.string, x = 0; x < (int) var->val_len; x++, cp++) {
+#if defined(_MSC_VER) /* Windows MSVC library includes non-ascii characters (above 0x7F) as printing, which can generate ugly strings */
```
bvanassche, line 650: One of the goals of the Net-SNMP project is to have the same behavior the same on all supported platforms. So I'm less than enthusiast about adding `#if defined(_MSC_VER)` here.

rgumpertz68:
I understand your hesitation.  This change actually makes the behavior under the MSVC library match more closely with other platforms that are more careful about what they treat as isprint.

I thought about adding the isascii call to all platforms, but what if the native character code isn’t ASCII?

What I submitted is the best compromise that I could think of.  Can you think of something better?

                              Rick

bvanassche:
How about this change (untested)?
```
diff --git a/snmplib/mib.c b/snmplib/mib.c
index 232c8756a9d4..07c4168d9247 100644
--- a/snmplib/mib.c
+++ b/snmplib/mib.c
@@ -647,7 +647,7 @@ sprint_realloc_octet_string(u_char ** buf, size_t * buf_len,
     case NETSNMP_STRING_OUTPUT_GUESS:
         hex = 0;
         for (cp = var->val.string, x = 0; x < (int) var->val_len; x++, cp++) {
-            if (!isprint(*cp) && !isspace(*cp)) {
+            if ((!isprint(*cp) || !isascii(*cp)) && !isspace(*cp)) {
                 hex = 1;
             }
         }
```

rgumpertz68:
Assuming that all you did was remove the #if, I’d say I can live with it.  Go ahead and do it…

                              Rick

PR #586, opened 2023-06-12
Fixed snmpv3 for multithreaded env via saving usm user in session for snmp_sess_open()

misha4gps:
This patch will help to use the library in a multi-threaded environment to access snmpv3 nodes. The main idea is to keep the user in the session and avoid calling usm_add_user() which is not thread-safe. The new mode is only available for the snmp_sess_open() function. All other behavior remains unchanged.

### DIFF
--- a/include/net-snmp/types.h
+++ b/include/net-snmp/types.h
@@ -418,6 +418,11 @@ struct snmp_session {
      * XXX: or should we add a new field into this structure?
      */
     void           *myvoid;
+
+    /**
+     * session specific user
+     */
+    struct usmUser *sessUser;
 };
 
 typedef struct netsnmp_index_s {
--- a/snmplib/snmpusm.c
+++ b/snmplib/snmpusm.c
@@ -501,8 +501,10 @@ usm_get_user_from_list(const u_char *engineID, size_t engineIDLen,
     /*
      * return "" user used to facilitate engineID discovery
      */
-    if (use_default && !strcmp(name, ""))
+    if (use_default && !strcmp(name, "")) {
+        DEBUGMSGTL(("usm", "return noNameUser\n"));
         return noNameUser;
+    }
     return NULL;
 }
 
@@ -1432,6 +1434,7 @@ usm_generate_out_msg(int msgProcModel,  /* (UNUSED) */
                       * secStateRef, pointer to cached info provided only for
                       * * Response, otherwise NULL.
                       */
+                     struct usmUser *sessUser,  /* IN - pointer to current session user. */
                      u_char * secParams,        /* OUT */
                      /*
                       * BER encoded securityParameters pointer to offset within
@@ -1533,7 +1536,11 @@ usm_generate_out_msg(int msgProcModel,  /* (UNUSED) */
          * we do allow an unknown user name for
          * unauthenticated requests. 
          */
-        user = usm_get_user2(secEngineID, secEngineIDLen, secName, secNameLen);
+        if (sessUser)
+            user = sessUser;
+        else
+            user = usm_get_user2(secEngineID, secEngineIDLen, secName, secNameLen);
+
         if (user == NULL && secLevel != SNMP_SEC_LEVEL_NOAUTH) {
             DEBUGMSGTL(("usm", "Unknown User(%s)\n", secName));
             return SNMPERR_USM_UNKNOWNSECURITYNAME;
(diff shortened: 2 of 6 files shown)


bvanassche (changes requested):

snmplib/snmpusm.c
```diff
@@ -1427,11 +1429,16 @@ usm_generate_out_msg(int msgProcModel,  /* (UNUSED) */
                       * * USM if needed.
                       */
                      size_t scopedPduLen,       /* IN - scopedPdu length. */
-                     const void *secStateRef, /* IN */
+                     const void *secStateRef,   /* IN */
```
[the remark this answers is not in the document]
misha4gps: Done

include/net-snmp/types.h
```diff
@@ -418,6 +418,11 @@ struct snmp_session {
      * XXX: or should we add a new field into this structure?
      */
     void           *myvoid;
+
```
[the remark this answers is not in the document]
misha4gps: Done.
Set LIBCURRENT  = 45

bvanassche:
Please combine the three commits into a single commit, e.g. as follows: `git reset HEAD^^^ && git commit -a && git push ... +master`.

misha4gps:
Done

snmplib/snmpusm.c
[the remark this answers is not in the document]
bvanassche: Hmm ... I still see a whitespace change?

snmplib/snmpusm.c
```diff
@@ -1427,11 +1429,16 @@ usm_generate_out_msg(int msgProcModel,  /* (UNUSED) */
...
                      /*
                       * secStateRef, pointer to cached info provided only for
                       * * Response, otherwise NULL.
                       */
+                     const u_long sessFlags,    /* IN */
```
bvanassche, line 1437: Why is this new argument needed? Isn't it sufficient to check whether or not the new sessUser argument is not NULL?

snmplib/snmpusm.c
```diff
@@ -3802,7 +3841,12 @@ usm_build_probe_pdu(netsnmp_pdu **pdu)
         user->privProtocolLen = OID_LENGTH(usmNoPrivProtocol);
         user->privProtocol =
             snmp_duplicate_objid(usmNoPrivProtocol, user->privProtocolLen);
-        usm_add_user(user);
+        if (session->flags & SNMP_FLAGS_SESSION_USER) {
```
bvanassche: Please add a patch at the start of this series that moves user creation out of usm_build_probe_pdu() such that 'user' is passed as an argument to usm_build_probe_pdu(). I think this is a cleaner solution rather than making usm_build_probe_pdu() modify session properties.

bvanassche (changes requested):

snmplib/snmpusm.c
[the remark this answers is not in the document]
misha4gps: No - removed in second commit

snmplib/snmpusm.c
misha4gps: Good point !
Done

snmplib/snmpusm.c
misha4gps: Done.
Should I combine all into one commit ?

bvanassche: Introducing changes in one commit and undoing these in another commit makes patches much harder to review than necessary. Please combine the "fixup" commit with the commit it fixes up.

misha4gps: I merged the latest changes to the previous commit.
Sorry for the question. Do you prefer to have all functional changes in the first commit and ABI only changes  left in another commit? (because some of the changes in the ABI commit were also functional)

bvanassche: Either way is fine.

The regression tests still fail. Please fix the regression test failures. The regression tests can be run locally as follows:
```
export MODE=regular && ci/build.sh && ci/net-snmp-run-tests
```

misha4gps: I fixed all tests except last 6 tests of 'perl' module related to tcp/udp/unixsocket. They are fail for me in master also.

bvanassche: It seems like the regression tests now pass?

Please combine the fixes patch into the first patch.

misha4gps: Done

bvanassche:
A slightly different patch has been applied (https://github.com/net-snmp/net-snmp/commit/1e4b822bb6690cf4548f26b661bb6708c64a1aa3). Please take a look.

misha4gps:
I agree `if (sessUser && *sessUser)` looks better
Thank you very much for your help!

PR #589, opened 2023-06-15
Declare our OpenSSL interface version.

nielsb:
This avoids warnings when compiling with OpenSSL 3.0+.

### DIFF
--- a/configure.d/config_net_snmp_config_h
+++ b/configure.d/config_net_snmp_config_h
@@ -42,6 +42,12 @@ AH_BOTTOM([/* end of definitions added by configure on-the-fly */
 #define HAVE_AES 1
 #endif
 
+#ifdef NETSNMP_USE_OPENSSL
+/* Suppress deprecation warnings from OpenSSL 3.0+ */
+/* Our interface match OpenSSL 1.0.0 */
+#define OPENSSL_API_COMPAT 0x10000000L
+#endif
+
 /* define signal if DNE */
 
 #ifndef HAVE_SIGNAL
--- a/include/net-snmp/net-snmp-config.h.in
+++ b/include/net-snmp/net-snmp-config.h.in
@@ -1974,6 +1974,12 @@
 #define HAVE_AES 1
 #endif
 
+#ifdef NETSNMP_USE_OPENSSL
+/* Suppress deprecation warnings from OpenSSL 3.0+ */
+/* Our interface match OpenSSL 1.0.0 */
+#define OPENSSL_API_COMPAT 0x10000000L
+#endif
+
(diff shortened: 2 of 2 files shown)


configure.d/config_net_snmp_config_h
```diff
@@ -42,6 +42,12 @@ AH_BOTTOM([/* end of definitions added by configure on-the-fly */
 #define HAVE_AES 1
 #endif
 
+#ifdef NETSNMP_USE_OPENSSL
+/* Suppress deprecation warnings from OpenSSL 3.0+ */
+/* Our interface match OpenSSL 1.0.0 */
+#define OPENSSL_API_COMPAT 0x10000000L
```
bvanassche, line 48: Will this definition end up in code that uses the Net-SNMP library? I'm wondering whether this definition should be moved into a private Net-SNMP header file.

nielsb: Yes, it will, as everything is including net-snmp-config.h, but I can't readily think of a possible candidate header for this?

bvanassche: I don't think that it is acceptable to add this code in the public header file so it may be necessary to create a new header file.

nielsb: Any idea how to name it?

bvanassche: How about `snmplib/openssl.h`? There are already several other private header files in that directory:
```
> ls snmplib/*h
snmplib/factory.h    snmplib/memcheck.h         snmplib/snmp_syslog.h
snmplib/inet_ntop.h  snmplib/snmpsm_init.h      snmplib/snprintf.h
snmplib/inet_pton.h  snmplib/snmpsm_shutdown.h  snmplib/ucd-snmp-includes.h
```

bvanassche: Isn't the `include/` directory for public header files? Shouldn't private header files occur outside the `include/` directory?

nielsb: There are a number of headers there already which are not installed:
data_access/ openssl_md5.h agent/hardware/ agent/object_monitor.h library/openssl_???.h library/snmp_openssl.h 
so it is nmot unheard of :-)

bvanassche: How about adding a new file in the include/net-snmp/library directory? It seems to me that the existing OpenSSL files in that directory come from the OpenSSL project so these files probably should not be modified. From commit 16fa55444e22af50be7eabbfbc71640d84285e7f: "local copies of the openssl headers".

PR #602, opened 2023-08-02
added support of asynchronous probing for USM instead of internal syn…

sokurenko:


bvanassche:
A slightly modified version of the provided patch has been applied as commit 9e0faeb24473. Please take a look.

PR #607, opened 2023-08-09
Add support for RPM SQLite DB background.

### LINKED ISSUE #596: Misleading error message for SQLite RPM db based systems
Hi,

starting from RPM 4.16 the SQLite support is available for RPM DB. After https://fedoraproject.org/wiki/Changes/Sqlite_Rpmdb, rpm changed from Berkeley DB to SQLite. SNMP is using these hard coded paths [1] to determine where RPM info is, so it is trying to search for /var/lib/rpm/Packages and if the system is running on SQLite DB for RPM it fails with the following error

```Can't find directory of RPM packages```

This should be handled as both options are possible in theory (depends on the system).

Question: Does it even make sense to keep this init check once no info is pulled from /var/lib/rpm/Packages directly, but everything is obtained via RPM API?

What do you think?

Regards 
Josef

[1] https://github.com/net-snmp/net-snmp/blob/master/agent/mibgroup/host/data_access/swinst_rpm.c#L82

jridky:
From RPM 4.16 the SQLite support is available for RPM DB. After https://fedoraproject.org/wiki/Changes/Sqlite_Rpmdb, rpm changed it's background DB from Berkeley to SQLite in Fedora. Net-SNMP is using hard coded paths to determine where RPM DB files are.

This update is adding check for rpmdb.sqlite file in order to be able invalidate internal cache after system package change.

Closes #596

### DIFF
--- a/agent/mibgroup/host/data_access/swinst_rpm.c
+++ b/agent/mibgroup/host/data_access/swinst_rpm.c
@@ -73,15 +73,23 @@ netsnmp_swinst_arch_init(void)
 #endif
 
     snprintf( pkg_directory, SNMP_MAXPATH, "%s/Packages", dbpath );
+    
+    if (-1 == stat( pkg_directory, &stat_buf )) {
+
+        /* check for SQLite DB backend */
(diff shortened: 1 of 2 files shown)


bvanassche:
Thanks for the patch!

PR #608, opened 2023-08-14
Handle SIGHUP for engine-id change

feroz-ahmed:
SNMP operations continue to work with old engine-id with SIGHUP.


feroz-ahmed:
I couldn't squash commits in my previous review, hence created a new one.

bvanassche:
This patch has been applied on the master branch. Thanks for the patch!

PR #614, opened 2023-08-16
Added tests to validate SIGHUP for exactEngineID change

feroz-ahmed:
This test validates commit 45a6422


bvanassche:
Thanks for the patch!

PR #616, opened 2023-08-18
mteTrigger.c: Fix copy-paste bug (Coverity 85524)

moshekaplan:
mteTrigger.c: Fix copy-paste bug (Coverity 85524)

Examine the proper variable when determining whether or not fire a fall event.


bvanassche:
Thanks for the patch!

PR #618, opened 2023-08-18
lookupResultsTable.c: Fix memory leaks (coverity 85588)

moshekaplan:
lookupResultsTable.c: Fix memory leaks (coverity 85588)

Free memory allocation referenced in StorageTmp that is not free()'d in the event of a failure


moshekaplan:
I noticed that pingCtlTable.c used `free` instead of `SNMP_FREE`. Would it be preferred to use that instead?

bvanassche:
I'd like to remove the SNMP_FREE() macro because it makes the source code harder to read than necessary.

moshekaplan:
Updated, thanks for your quick response.

bvanassche:
Please combine the two commits into a single commit.

moshekaplan:
Squashed

bvanassche:
Thanks for the patch!

PR #619, opened 2023-08-18
pingCtlTable.c: Fix memory leaks (Coverity 85603)

moshekaplan:
pingCtlTable.c: Fix memory leaks (Coverity 85603)

Free allocated memory if an error occurs.

### DIFF
--- a/agent/mibgroup/disman/ping/pingCtlTable.c
+++ b/agent/mibgroup/disman/ping/pingCtlTable.c
@@ -418,6 +418,7 @@ parse_pingCtlTable(const char *token, char *line)
                               &StorageTmp->pingCtlOwnerIndexLen);
     if (StorageTmp->pingCtlOwnerIndex == NULL) {
         config_perror("invalid specification for pingCtlOwnerIndex");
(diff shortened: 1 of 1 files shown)


agent/mibgroup/disman/ping/pingCtlTable.c
```diff
@@ -4223,7 +4230,7 @@ write_pingCtlRowStatus(int action,
             vars = NULL;
 
             /*
-             * ��nameΪ�յ����������ֶμӵ�var�����б���ĩβ 
+             * 将name为空的三个索引字段加到var变量列表的末尾
```
bvanassche, line 4233: This looks like an unintended change?

moshekaplan: You're right. My guess is that the Github web editor added that. I'll figure out a way to remove it.

moshekaplan: Should be good now. Thanks for catching that accidental modification.

bvanassche:
Thanks for the patch!

PR #621, opened 2023-08-21
snmpNotifyFilterProfileTable_data.c: Fix memory leaks (Coverity 85636)

moshekaplan:
snmpNotifyFilterProfileTable_data.c: Fix memory leaks (Coverity 85636)

Free memory allocation referenced in StorageTmp that is not free()'d in the event of a failure


bvanassche:
Thanks for the patch!

PR #622, opened 2023-08-21
icmp.c: Fix memory leaks (Coverity 85667)

moshekaplan:
icmp.c: Fix memory leaks (Coverity 85667)

Free memory allocation referenced in msg_stats_table_info that is not free()'d in the event of a failure


bvanassche:
Thanks for the patch!

PR #623, opened 2023-08-21
nsDebug.c: Fix memory leak (Coverity 85684)

moshekaplan:
nsDebug.c: Fix memory leak (Coverity 85684)

Free memory allocation referenced in table_info that is not free()'d in the event of a failure


bvanassche:
Thanks for the patch!

PR #624, opened 2023-08-21
expValue.c: Fix memory leak (Coverity 85691)

moshekaplan:
expValue.c: Fix memory leak (Coverity 85691)

Avoid leaking memory referenced in var if invalid parameters are passed in by only allocating memory if parameters are valid.


bvanassche:
Thanks for the patch!

bvanassche:
Thanks for the patch!

PR #625, opened 2023-08-21
lookupCtlTable.c: Fix memory leak (Coverity 85698-4)

moshekaplan:
Avoid leaking memory referenced in address if an invalid argument is passed in.


bvanassche:
Thanks for the patch!

PR #626, opened 2023-08-21
table_dataset.c: Fix memory leak (Coverity 144371)

moshekaplan:
Free memory allocation referenced in table_info that is not free()'d in the event of a failure


bvanassche:
Thanks for the patch!

PR #627, opened 2023-08-21
deliverByNotify.c: Fix memory leak (Coverity 144379)

moshekaplan:
Avoid leaking memory referenced in vars if it fails to issue the query.


bvanassche:
Thanks for the patch!

PR #628, opened 2023-08-21
route_linux.c: Fix memory leak (Coverity 344582)

moshekaplan:
route_linux.c: Fix memory leak (Coverity 344582)

Avoid leaking memory referenced in entry if the IPv6 data is formatted incorrectly.


bvanassche:
Thanks for the patch!

PR #629, opened 2023-08-21
netSnmpHostsTable.c: Fix memory leak (Coverity 344583)

moshekaplan:
netSnmpHostsTable.c: Fix memory leak (Coverity 344583)

Free memory allocation referenced in my_handler that is not free()'d in the event of a failure.

### DIFF
--- a/agent/mibgroup/examples/netSnmpHostsTable.c
+++ b/agent/mibgroup/examples/netSnmpHostsTable.c
@@ -65,8 +65,9 @@ initialize_table_netSnmpHostsTable(void)
     if (!my_handler || !table_info || !iinfo) {
         snmp_log(LOG_ERR,
                  "malloc failed in initialize_table_netSnmpHostsTable");
-        free(iinfo);
+        free(my_handler);
         free(table_info);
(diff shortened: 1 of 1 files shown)


agent/mibgroup/examples/netSnmpHostsTable.c
```diff
@@ -65,8 +65,15 @@ initialize_table_netSnmpHostsTable(void)
     if (!my_handler || !table_info || !iinfo) {
         snmp_log(LOG_ERR,
                  "malloc failed in initialize_table_netSnmpHostsTable");
-        free(iinfo);
-        free(table_info);
+        if (my_handler) {
+            free(my_handler);
+        }
+        if (iinfo) {
```
bvanassche, line 71: Please leave out the if-tests around the free() calls. From https://pubs.opengroup.org/onlinepubs/007904975/functions/free.html: _If ptr is a null pointer, no action shall occur._

moshekaplan:
Note: I reordered the free(iinfo) to mirror the if statement; it seemed like it would make it easier to read.

PR #632, opened 2023-08-24
icmp.c: Avoid uninitialized variable error

moshekaplan:
Set the variable msg_stats_table_info to a default value of NULL, to avoid potentially accessing the pointer when it's uninitialized

### DIFF
--- a/agent/mibgroup/mibII/icmp.c
+++ b/agent/mibgroup/mibII/icmp.c
@@ -638,7 +638,7 @@ init_icmp(void)
     netsnmp_iterator_info *iinfo;
     netsnmp_iterator_info *msg_stats_iinfo;
     netsnmp_table_registration_info *table_info = NULL;
-    netsnmp_table_registration_info *msg_stats_table_info;
+    netsnmp_table_registration_info *msg_stats_table_info = NULL;
 #endif
     netsnmp_handler_registration *scalar_reginfo = NULL;
     int                    rc;


agent/mibgroup/mibII/icmp.c
```diff
@@ -638,7 +638,7 @@ init_icmp(void)
     netsnmp_iterator_info *iinfo;
     netsnmp_iterator_info *msg_stats_iinfo;
     netsnmp_table_registration_info *table_info = NULL;
-    netsnmp_table_registration_info *msg_stats_table_info;
+    netsnmp_table_registration_info *msg_stats_table_info = NULL;
```
bvanassche, line 641: Hmm ... I don't see any code path that reads this pointer before it has been set?

moshekaplan: I too did not see how it would be triggered, but gcc was complaining:

mibgroup/mibII/icmp.c: In function ‘init_icmp’:
mibgroup/mibII/icmp.c:754:8: warning: ‘msg_stats_table_info’ may be used uninitialized [-Wmaybe-uninitialized]
  754 |     if (msg_stats_table_info)
      |        ^
mibgroup/mibII/icmp.c:641:38: note: ‘msg_stats_table_info’ was declared here
  641 |     netsnmp_table_registration_info *msg_stats_table_info ;
      |                                      ^~~~~~~~~~~~~~~~~~~~

It seemed like initializing it to NULL would be less risky than ignoring the warning. And given that I introduced the 'bug' in the first place, I didn't want to leave a mess behind.

bvanassche: Now I see what's going on: `free(msg_stats_table_info)` can be executed before `msg_stats_table_info` has been initialized. I will merge this patch.

PR #633, opened 2023-08-25
snmpUnixDomain.c: Free memory if recvfrom fails

moshekaplan:
Free memory if recvfrom() fails.
Found with clang scan-build


snmplib/transports/snmpUnixDomain.c
```diff
@@ -150,6 +150,9 @@ netsnmp_unix_recv(netsnmp_transport *t, void *buf, int size,
             if (rc < 0 && errno != EINTR) {
                 DEBUGMSGTL(("netsnmp_unix", "recv fd %d err %d (\"%s\")\n",
                             t->sock, errno, strerror(errno)));
+                free(to);
+                *opaque = NULL;
+                *olength = 0;
```
bvanassche, line 155: The caller should not access `*opaque` nor `*olength` if a negative value is returned so I think that adding `free(to)` is sufficient.

moshekaplan: Changed to that.

PR #634, opened 2023-08-25
snmpv3.c: Confirm snmpv3_generate_engineID() was successful

moshekaplan:
Confirm snmpv3_generate_engineID() was successful before using data from it

If snmpv3_generate_engineID() fails, it returns NULL, possibly without modifying the passed-in engineIDLen parameter. This change examines the return value from snmpv3_generate_engineID() before examining the possibly-uninitialized engineIDLen.

Found with clang scan-build


bvanassche:
Thanks for the patch!

PR #636, opened 2023-08-27
lookupCtlTable.c: Fix memory leak (Coverity 85698-2)

moshekaplan:
lookupCtlTable.c: Fix memory leak (Coverity 85698-2) Avoid leaking memory referenced in address if IPv4 DNS lookup fails.


bvanassche:
Thanks for the patch!

PR #637, opened 2023-08-27
lookupCtlTable.c: Fix memory leak (Coverity 85698-3)

moshekaplan:
lookupCtlTable.c: Fix memory leak (Coverity 85698-3) Avoid leaking memory referenced in address if IPv6 DNS lookup fails.


bvanassche:
Thanks for the patch!

PR #638, opened 2023-08-27
lookupCtlTable.c: Fix memory leak (Coverity 85732)

moshekaplan:
Avoid leaking memory referenced in StorageTmp if invalid data is passed in.


bvanassche:
Thanks for the patch!

PR #639, opened 2023-08-27
subagent.c: Fix memory leak (Coverity 357928)

moshekaplan:
Avoid leaking memory referenced in smagic if snmp_clone_pdu fails.


bvanassche:
Thanks for the patch!

PR #641, opened 2023-08-28
read_config.c: Remove dead store

moshekaplan:
Remove dead store detected by clang.

read_config.c:1454:26: warning: Value stored to 'ctmp' during its initialization is never read [deadcode.DeadStores]
    struct config_files *ctmp = config_files;


moshekaplan:
Please let me know if such PRs are desired

bvanassche:
Thanks for the patch! I think removing dead stores is fine since this reduces the number of warnings reported by clang.

PR #643, opened 2023-08-28
snmpTlstmCertToTSNTable.c: Fix memory leak (Coverity 85750)

moshekaplan:
snmpTlstmCertToTSNTable.c: Fix memory leak (Coverity 85750)

Avoid leaking memory referenced in watcher if netsnmp_create_handler_registration fails.

### DIFF
--- a/agent/mibgroup/tlstm-mib/snmpTlstmCertToTSNTable/snmpTlstmCertToTSNTable.c
+++ b/agent/mibgroup/tlstm-mib/snmpTlstmCertToTSNTable/snmpTlstmCertToTSNTable.c
@@ -1322,16 +1322,20 @@ init_snmpTlstmCertToTSNTable_context(const char *contextName)
     }
     
     reg_oid[10] = 2;
-    to_tsn_last_changed_reg = netsnmp_create_handler_registration(
-        "snmpTlstmCertToTSNTableLastChanged", NULL, reg_oid,
-        OID_LENGTH(reg_oid), HANDLER_CAN_RONLY);
     watcher = netsnmp_create_watcher_info((void*)&_last_changed,
                                           sizeof(_last_changed),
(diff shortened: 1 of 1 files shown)


agent/mibgroup/tlstm-mib/snmpTlstmCertToTSNTable/snmpTlstmCertToTSNTable.c
```diff
@@ -1322,16 +1322,20 @@ init_snmpTlstmCertToTSNTable_context(const char *contextName)
     }
     
     reg_oid[10] = 2;
-    to_tsn_last_changed_reg = netsnmp_create_handler_registration(
```
bvanassche, line 1325: It is not clear to me why the order of the netsnmp_create_handler_registration() and netsnmp_create_watcher_info() calls has been changed?

moshekaplan: Because `to_tsn_last_changed_reg` begins as uninitialized and is only set if the `watcher` was successfully created

moshekaplan: My apologies, I misread your question.
The reason why I switched the ordering of the calls to `netsnmp_create_handler_registration()` and `netsnmp_create_watcher_info()`  is because it seemed like both `to_tsn_last_changed_reg` and `watcher` are both required for the next step and that `watcher` would be the easier of the two variables to clean up if `netsnmp_create_handler_registration()` failed, because it's only a memory allocation and does not modify any external variables.

bvanassche:
Thanks for the patch!

PR #648, opened 2023-08-29
Makefile.top: Use --tag=CC when invoking libtool

### LINKED ISSUE #647: libtool: compile: unable to infer tagged configuration
When I configure net-snmp 5.9.4 (on macOS 12.6.7), I set the CC environment variable to my preferred compiler, e.g.:

```
CC=/usr/bin/clang ./configure
```

This works except that the perl modules are built using the compiler that was used to build perl, not the compiler I requested.

To work around this, I additionally set CC and LD as arguments when building, e.g.:

```
make CC=/usr/bin/clang LD=/usr/bin/clang
```

However, because the compiler I specified is not the one I compiled libtool with, this fails with:

```
libtool: compile: unable to infer tagged configuration
libtool:   error: specify a tag with '--tag'
```

The solution is as the error message says: Always specify the tag when invoking libtool (with `--mode=compile` or `--mode=link`). I'll submit a PR to fix this.

ryandesign:
Always use `--tag=CC` when invoking libtool with `--mode=compile` or `--mode=link` to avoid the error message "libtool: compile: unable to infer tagged configuration" when compiling with a different compiler than the one that was used to compile libtool.

Fixes #647

### DIFF
--- a/Makefile.top
+++ b/Makefile.top
@@ -86,18 +86,18 @@ LIBCURRENT  = 45
 LIBAGE      = 0
 LIBREVISION = 0
 
-LIB_LD_CMD      = $(LIBTOOL) --mode=link $(LINKCC) $(CFLAGS) $(LDFLAGS) -rpath $(libdir) -version-info $(LIBCURRENT):$(LIBREVISION):$(LIBAGE) @LD_NO_UNDEFINED@ -o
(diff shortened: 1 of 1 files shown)


bvanassche:
Please provide a better description than "Fixes #647". Patch descriptions should be comprehensible without having to follow hyperlinks.

ryandesign:
Is this better? Feel free to edit my commit message further if needed.

bvanassche:
Thanks for the patch!

PR #649, opened 2023-08-30
snmplib: do not filter internal request

lioyerar:
sourceFilterType and sourceFilterAddress options in snmpd.conf. However, once activated, internal requests are also blocked. The filtering is based on IP in the string returned by the function "fmtaddr". Internal requests do not have a source IP, instead of an IP address, the function returns a string that starts with "callback" following a file descriptor numero. These internal requests are generated by alarms to define event triggers. If they are blocked, no traps will be sent.

Therefore, do not filter internal requests by checking if the string returned by "fmtaddr" starts with "callback".

Fixes: 9d28612ac03b ("add packet filtering by source ip")


bvanassche:
Thanks for the patch!

Issue #650, opened 2023-08-30
Missing semicolon causes build error


jridky:
Hi,

I received report from s390utils maintainer, that because of missing semicolon, the build of this package is failing.

From further investigation, I've found out, that issue is in the file

https://github.com/net-snmp/net-snmp/blob/e1e0c9dfe5b54e81e4db8fa63226bcf5de206c03/configure.d/config_modules_agent#L30

where function call should end with semicolon.

Origin of this mistake is from commit a2cb167.

Fix will be delivered ASAP. Tested with s390utils on Fedora Rawhide and works fine.

jridky:
The issue was in s390utils so I proposed the fix there instead.

Thanks for the review.

PR #651, opened 2023-08-30
Add missing semicolon to function definition

jridky:
It was discovered, that missing semicolon causes build failure of s390utils package. From further investigation it shows that function call netsnmp_unused_function(void) should be ended by semicolon.

This issue was introduced via commit a2cb167 and reported in issue #650.

### DIFF
--- a/configure.d/config_modules_agent
+++ b/configure.d/config_modules_agent
@@ -27,7 +27,7 @@ done
 #
 AC_MSG_CHECKING([for and configuring mib modules to use])
 
-AH_TOP([#define NETSNMP_REQUIRE_SEMICOLON extern void netsnmp_unused_function(void)])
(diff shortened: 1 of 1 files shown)


configure.d/config_modules_agent
```diff
@@ -27,7 +27,7 @@ done
 #
 AC_MSG_CHECKING([for and configuring mib modules to use])
 
-AH_TOP([#define NETSNMP_REQUIRE_SEMICOLON extern void netsnmp_unused_function(void)])
+AH_TOP([#define NETSNMP_REQUIRE_SEMICOLON extern void netsnmp_unused_function(void);])
```
bvanassche, line 30: This change is wrong and breaks the NETSNMP_REQUIRE_SEMICOLON macro.

Issue #652, opened 2023-08-30
Regression in agent/mibgroup/host/hr_swinst.c


jridky:
Hi,

I've noticed a memory leak regression in agent/mibgroup/host/hr_swinst.c file.

There was proposed patch in PR #49 that solves issue with memory leak, when using RPMDB macros to read config file path.

Unfortunatelly, this change didn't manage it to become part of net-snmp-5.9.4 and was removed during the time (not sure when exactly).

Reimplementing of this fix will be delivered in the new PR.

PR #653, opened 2023-08-30
Fix regression of memory leak when using RPMDB macros.

jridky:
This change is fixing memory leak when using RPMDB macros. The fix was introduced in PR #49 but was overwritten over time. Fix is needed in agent/mibgroup/host/hr_swinst.c only as the regression in the agent/mibgroup/host/data_access/swinst_rpm.c was fixed few weeks ago.

This is proposed solution for issue #652


bvanassche:
Thanks for the patch!

PR #655, opened 2023-08-30
old_api.c: Check result of netsnmp_duplicate_variable()

moshekaplan:


bvanassche:
Thanks for the patch!

PR #656, opened 2023-08-30
master_admin.c: Validate that init_agent_snmp_session() was successful

moshekaplan:
master_admin.c: Validate that init_agent_snmp_session() was successful


bvanassche:
Thanks for the patch!

PR #658, opened 2023-08-30
container_binary_array.c: Validate that netsnmp_binary_array_initiali…

moshekaplan:
…ze() was successful


bvanassche:
Thanks for the patch!

PR #659, opened 2023-08-30
parse.c: Validate that strdup() was successful

moshekaplan:


bvanassche:
Thanks for the patch!

PR #661, opened 2023-08-30
snmp_enum.c: Validate that strdup() was successful

moshekaplan:


moshekaplan:
Note: In PR https://github.com/net-snmp/net-snmp/pull/404 , the condition for allocation was modified to `if (!created && list) `

bvanassche:
Thanks for the patch!

PR #662, opened 2023-08-30
snmp_service.c: Validate that malloc() was successful

moshekaplan:


bvanassche:
Thanks for the patch!

PR #663, opened 2023-08-30
snmp_transport.c: Validate memory allocations

moshekaplan:
Ensure malloc() and strdup() were successful


bvanassche:
Thanks for the patch!

PR #664, opened 2023-08-30
tools.c: Ensure memory allocations were successful

moshekaplan:


bvanassche:
Thanks for the patch!

PR #665, opened 2023-08-31
snmpd.conf.5.def: Fix misspellings and grammar typos

moshekaplan:


bvanassche:
Thanks for the patch!

PR #666, opened 2023-08-31
agentxtrap.1.def: Typos and grammar

moshekaplan:


bvanassche:
Thanks for the patch!

PR #667, opened 2023-08-31
encode_keychange.1.def: Typo

moshekaplan:


bvanassche:
Thanks for the patch!

PR #668, opened 2023-08-31
mib2c-update.1.def: Typos

moshekaplan:


bvanassche:
Thanks for the patch!

PR #669, opened 2023-08-31
mib2c.1.def: Hyphens, missing word

moshekaplan:


bvanassche:
Thanks for the patch!

PR #670, opened 2023-08-31
mib2c.conf.5.in: Grammar

moshekaplan:


bvanassche:
Thanks for the patch!

PR #674, opened 2023-08-31
pingCtlTable.c: Validate create_pingCtlTable_data() allocations

moshekaplan:
Ensure all memory allocations within create_pingCtlTable_data() were successful


bvanassche:
Thanks for the patch!

PR #682, opened 2023-09-07
Cache handler misc fixes

Ilyesbenhamouda:


agent/helpers/cache_handler.c
```diff
@@ -785,13 +788,21 @@ release_cached_resources(unsigned int regNo, void *clientargs)
...
             } else {
                 cache_outstanding_valid = 1;
             }
         }
     }
+
+    if (do_trim)
+        /* Release freed memory back to the system */
+        malloc_trim(0);
+
```
bvanassche, line 805: All systems I know of that run Net-SNMP support virtual memory. Is calling malloc_trim() on such systems useful?

Ilyesbenhamouda: Yes, the patch was actually tested on Linux.

bvanassche:
Thanks for the patches!

PR #683, opened 2023-09-10
snmpget.c: Avoid leak if parsing OID fails

moshekaplan:
snmpget.c: Avoid leak if parsing OID fails. This leak would not normally cause any problems, but might make investigating other leaks more difficult


bvanassche:
Thanks for the patch!

PR #684, opened 2023-09-10
parse_octet_hint_fuzzer.c: Use size_t for value_start

moshekaplan:
Fixes compiler warning:

Compiling testing/fuzzing/parse_octet_hint_fuzzer.c testing/fuzzing/parse_octet_hint_fuzzer.c:58:24: error: comparison of integers of different signs: 'int' and 'size_t' (aka 'unsigned long') [-Werror,-Wsign-compare]
    assert(value_start <= size);

Everything else is a size_t, no reason to not have value_start as a size_t too


bvanassche:
Please note that there are no plans to enable `-Wsign-compare`.

PR #706, opened 2023-09-19
Adding server cipher order by default

NicoBusuioc:
According to testssl tool (https://testssl.sh/), it is not ok when Server has not set a cipher order, see image below

image


bvanassche:
Thanks for the patch!

PR #710, opened 2023-09-21
V5 9 patches: Another set of backports from master

nielsb:
Ever thought of applying to V5-9-patched instead of master?


bvanassche:
> Ever thought of applying to V5-9-patched instead of master?

I think we should have a single branch for Net-SNMP instead of two branches.

bvanassche:
On 9/21/23 11:14, Niels Baggesen wrote:
> Ever thought of applying to V5-9-patched instead of master?

An additional concern is that the creator of a pull request selects the
target branch. The github user interface does not allow to change the
target branch.

PR #714, opened 2023-09-26
Spelling and grammar fixes

moshekaplan:
Mostly found with:
`find . -type f -name "*.c" -exec aspell --mode=ccpp check {} \;`

Common replacements made with:
`grep -rl 'writable' ./ | xargs sed -i 's/writeable/writable/g'`


fenner (changes requested):
Did you miss the `commited` examples in mib2c-conf.d/parent-set.m2i and mfd-interface.m2c?

There are several instances of `writeable` in local/ in net-snmp-cert and in mib2c-conf.d

In general the source of these repeated consistent typos is a typo in the templates, so updating the templates would help prevent future instances

moshekaplan:
@fenner : I did not review all files. That's why there are probably still plenty of typos remaining. I figured that some improvement is better than nothing :)

agent/mibgroup/Rmon/rows.c
```diff
@@ -67,7 +67,7 @@ rowapi_delete(RMON_ENTRY_T * eold)
     }
 
     /*
-     * delete it in users's sence 
+     * delete it in users's sense
```
bvanassche, line 70: Although I'm not a native speaker: this doesn't look right to me.

moshekaplan: sense is probably correct. It should be user's though.

moshekaplan: Added as an additional commit.

bvanassche:
Thanks for the patches!

PR #718, opened 2023-10-03
Enable IPV6_RECVPKTINFO

xelkyr:


configure.d/config_os_misc4
```diff
@@ -369,6 +369,56 @@ if test x$ac_cv_have_ip_pktinfo = xyes; then
     AC_DEFINE(HAVE_IP_PKTINFO, 1, [Set if IP_PKTINFO is usable])
 fi
 
+#   Check whether IPV6_RECVPKTINFO is usable.
+#
+
+AC_CACHE_CHECK([for IPV6_RECVPKTINFO ],
+                [ac_cv_have_ipv6_recvpktinfo],
+[AC_LINK_IFELSE([AC_LANG_PROGRAM([
```
bvanassche, line 377: Since this program is compiled and linked but not run, please replace it with the following:
```
return IPV6_RECVPKTINFO
```

snmplib/transports/snmpUDPIPv6Domain.c
```diff
@@ -288,6 +503,16 @@ netsnmp_udp6_transport_bind(netsnmp_transport *t,
             } 
         }
 #endif
+#if defined(HAVE_IPV6_RECVPKTINFO) && !defined(WIN32)
```
bvanassche, line 506: Please move this code, including the #if / #endif, into a helper function.

bvanassche (changes requested):

configure.d/config_os_misc4
xelkyr: Updated, thank you.

snmplib/transports/snmpUDPIPv6Domain.c
xelkyr: Moved it to helper function right above.

bvanassche:
Thanks for the patches!

PR #719, opened 2023-10-05
snmptrap.c: Properly clean up session values

moshekaplan:
snmptrap.c: Properly clean up session values
Avoid zero-ing session structure so that its members can be free()'d.

### DIFF
--- a/apps/snmptrap.c
+++ b/apps/snmptrap.c
@@ -378,7 +378,6 @@ main(int argc, char *argv[])
 
 close_session:
     snmp_close(ss);
-    memset(&session, 0, sizeof(session));
     snmp_shutdown(NETSNMP_APPLICATION_CONFIG_TYPE);
 
 out:


moshekaplan:
@bvanassche : It was unclear to me why `memset(&session, 0, sizeof(session));` was ever present at all, is it possible it was simply a holdover from before `netsnmp_cleanup_session(&session);` was added?

bvanassche:
I think this patch does not fix any memory leaks and also that it *introduces* a use-after-free. snmp_close() frees all the memory owned by the `session` object.

bvanassche:
Please take a look at commits 44abb16846f742bdbc7541de3f87e93105d4a50e and bc15ada7adc35011596f143672057bb8fce345ea.

moshekaplan:
Those commits appear to solve the memory leaks and so this PR is no longer needed. Thank you!

PR #729, opened 2023-10-12
snmpusm.c: Fix segfault when parsing config file createUser line

### LINKED ISSUE #416: Crash when parsing config (createUser)
Version: `v5.9.1`
Configure options: `-without-perl-modules --disable-embedded-perl --with-cflags=-g  --disable-mib-loading`

Reproduce: add this line to your _**snmpd.conf**_
`createUser test blah`
This is a bad configuration, and it should result in error instead of crashing

Call stack:
`read_config() -> run_config_handler() -> usm_parse_create_usmUser() -> usm_create_usmUser_from_string()`
at: `usm_create_usmUser_from_string()`
in: api/net-snmp/snmplib/snmpusm.c
line: `4489`
code: `if ((strncmp(cp, "default", 7) == 0) && (NULL != def_auth_prot)) {`
root cause: **_cp_** is 0x0, which causes strncmp to crash

Suggestion: change the line to
`if (( cp && strncmp(cp, "default", 7) == 0) && (NULL != def_auth_prot)) {`

BTW: I couldn't build the latest(master) branch, therefore I filed an issue (instead of fixing it myself)
error:
`mibgroup/if-mib/data_access/interface_linux.c:27:2: error: #error libnl-3 is required. Please install the libnl-3 and libnl-route-3 development packages and remove --without-nl from the configure options if necessary.`
I've installed `libnl-*-3-dev` on `Ubuntu 20.04`, reconfigured, make clean, and make again. No luck.
Perhaps I'm missing something, I'd appreciate it if anyone can point me to the latest development guide.

Thanks.

moshekaplan:
snmpusm.c: Fix segfault when parsing config file createUser line If no authentication profile is specified, use the default.

Fixes #416

### DIFF
--- a/snmplib/snmpusm.c
+++ b/snmplib/snmpusm.c
@@ -4674,7 +4674,10 @@ usm_create_usmUser_from_string(char *line, const char **errorMsg)
      */
     newuser->authProtocol[0] = 0;
     cp = copy_nword(cp, buf, sizeof(buf));
-    if ((strncmp(cp, "default", 7) == 0) && (NULL != def_auth_prot)) {
+    /* If no authentication protocol was specified, or it was explicitly
+     * set to use the default, use the default auth protocol
+     */
+    if ((!cp || (strncmp(cp, "default", 7) == 0)) && (NULL != def_auth_prot)) {
         SNMP_FREE(newuser->authProtocol);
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #740, opened 2023-10-17
snmp_transport.c: Remove redundant if statement

moshekaplan:
Found with cppcheck


bvanassche:
Thanks for the patch!

bvanassche:
Thanks for the patch!

PR #741, opened 2023-10-20
coverity.yml: Only run in main repository

moshekaplan:
This prevents the Coverity scheduled job from automatically running in forked repos

`github.repository` is documented here: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

YAML entry is similar to here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -10,6 +10,7 @@ on:
   workflow_dispatch:
 jobs:
   coverity:
+    if: github.repository == 'net-snmp/net-snmp'
     runs-on: ubuntu-latest
     steps:
     - name: Check github variables


.github/workflows/coverity.yml
```diff
@@ -10,6 +10,7 @@ on:
   workflow_dispatch:
 jobs:
   coverity:
+    if: github.repository == 'net-snmp/net-snmp'
```
bvanassche, line 13: Please change this into a test that verifies that both secrets.COVERITY_SCAN_EMAIL and secrets.COVERITY_SCAN_TOKEN have been set. These keys have been set on my github repository but not yet on the official github repository. I plan to ask Wes soon to set these secrets on the official github repository.

bvanassche (changes requested):

.github/workflows/coverity.yml
moshekaplan: I googled this for a bit and unfortunately, GitHub does not appear to support a job-level conditional to only run a job when a secret is set.

bvanassche: That's unfortunate. Anyway, thanks for having looked this up.

bvanassche:
Thanks for the patch!

.github/workflows/coverity.yml
bvanassche, line 13: Does this test work as intended? The coverity.yml action still runs on my own github repository as one can see here: https://github.com/bvanassche/net-snmp/actions/runs/6647844856/job/18063944527

moshekaplan:
I am looking at the master branch in your folk and don't see the if
statement in it:
https://github.com/bvanassche/net-snmp/blob/master/.github/workflows/coverity.yml

Am I missing something?

bvanassche:
This has been fixed by updating my copy of the Net-SNMP repository.Message
ID: ***@***.***>

>

PR #743, opened 2023-10-30
V5 9 patches

nielsb:
Backporting current bug fixes from master


bvanassche:
Thank you for having backported these patches.

PR #748, opened 2023-11-22
Give memory back to system post snmp operation

feroz-ahmed:
On low-end devices, there is no point to hold the memory with sub-agent. 
This memory can be used by other process.


feroz-ahmed:
Do we see any issue with this review?

bvanassche:
What is the performance impact of this patch? How much does it slow down Net-SNMP?

feroz-ahmed:
We are calling malloc_trim() from CONTAINER_CLEAR() which gets called from _container_free().
_container_free() , gets called on cache-timer expiry.

So we havent see any performance impact, in snmpwalk/get operations.

PR #749, opened 2023-11-24
Update V5-90-patches with my fixes for modern OpenSSL

nielsb:
I hope someone will pick it up and continue.
There are now 5-7 failures when running the TLS test suite, if you run without leak checking.
There are a number of leaks that still needs fising too.


bvanassche:
Thanks for the patches!

PR #755, opened 2023-12-13
Perl: added deallocation of session in error handling path

ProjectMutilation:
Found by RASU JSC


bvanassche:
Thanks for the patch!

PR #756, opened 2023-12-13
Protect malloc_trim() call with configure #ifdef

nielsb:
This fixes 460a97c  "Give memory back to system post snmp operation" for BSD and others without the malloc_trim() function.


bvanassche:
Thanks for the patch!

PR #759, opened 2023-12-14
agent: fixed memory leak

ProjectMutilation:
Macro netsnmp_table_row_add_index called function
snmp_varlist_add_variable() that allocated memory by calling macro SNMP_MALLOC_TYPEDEF and returned pointer to this memory
Added assigment and call of free() for avoiding leaking  

Found by RASU JSC
Signed-off-by: Maxim Korotkov <maskorotkov@rasu.ru>

### DIFF
--- a/agent/helpers/table_dataset.c
+++ b/agent/helpers/table_dataset.c
@@ -1173,7 +1173,7 @@ netsnmp_config_parse_add_row(const char *token, char *line)
     int             rc;
 
     data_set_tables *tables;
-    netsnmp_variable_list *vb;  /* containing only types */
+    netsnmp_variable_list *vb, *vars;  /* containing only types */
     netsnmp_table_row *row;
     netsnmp_table_data_set_storage *dr;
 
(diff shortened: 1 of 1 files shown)


ProjectMutilation:
@bvanassche hi, can you give any feedback? just friendly reminder

bvanassche:
Please fix the RASU JSC software. I think that the patch in this pull request introduces a dangling pointer and also that there is no leak in the modified code.

ProjectMutilation:
I found this when I was just researching the code, it seemed somewhat strange to me that the function that returns the pointer is used as void, I think that if `*varlist = vars;` fails, a leak is possible

I do not exclude the possibility that I am wrong

bvanassche:
> I found this when I was just researching the code, it seemed somewhat strange to me that the function that returns the pointer is used as void, I think that if `*varlist = vars;` fails, a leak is possible

Hmm ... how can `*varlist = vars;` fail?

ProjectMutilation:
This statement in condition, i told about it, but i agree with you and this pull request is unneeded

PR #761, opened 2023-12-19
agent: mibgroup: avoid potential NULL dereference

ProjectMutilation:
The "while" loop shifts the pointer,
which can cause it to be NULL.
Fixes: 8baf64d4ae616(Re-worked Schedule MIB implementation.)  
Found by RASU JSC


ProjectMutilation:
@bvanassche hi, just a quick reminder, what do you think of the patch?

bvanassche:
Thanks for the patch!

PR #764, opened 2023-12-23
Add 'ubifs' to other_fs

chleroy:
ubifs used to be automatically taken into account in UCD-SNMP-MIB::dskTable but commit e3fc76e0ae ("CHANGES: snmpd: Make UCD-SNMP::dskTable dynamic if includeAllDisks is set.") added a verification that drops all filesystems not present in other_fs[] table.

So add 'ubifs' in other_fs[] to fix it.


bvanassche:
Thanks for the patch!

PR #770, opened 2024-01-09
V5-9-patches

nielsb:
Fixing warnings, mostly OpenBSD, NetBSD.


bvanassche:
Thanks for the patches!

Issue #774, opened 2024-01-16
Timeout upon unknown context name


lukistar14:
Hello net-snmp experts,

I found that snmpd gives no response when an unknown context name is given in the request. The client subsequently runs into a timeout.
There is a comment at the relevant portion of the code, stating (agent/snmp_agent.c):

> rfc3413 section 3.2, step 5 says that we increment the
>  counter but don't return a response of any kind 

Here is the respective paragraph:

> If the isAccessAllowed ASI returns a noSuchContext error,
> processing of the management operation is halted, no result PDU
> is generated, the snmpUnknownContexts counter is incremented,
> and control is passed to step (6) below for generation of a
> report message.

See [RFC3413](https://datatracker.ietf.org/doc/html/rfc3413#section-3.2) for the whole section.

IMO this means that a report message should be sent, or am I missing something?

PR #776, opened 2024-01-17
Report unknown context

lukistar14:
So far, when the client uses an unknown context in a request, nothing is responded and the request runs into a timeout. Now we report the snmpUnknownContexts counter.

References issue #774. Not sure whether the behaviour introduced here is snmp spec conformant; any input is appreciated.


hardaker:
Makes logical sense, but we need to see what the SNMPv3 RFCs say first.  Sometimes SNMP deliberately doesn't respond to avoid probing issues.

PR #782, opened 2024-01-30
mib2c: Fix re-initialization of tables

SteinHeselmans:
Allows re-initialization of tables, if e.g. a table needs to be registered under a different snmp-context or to a different agent on a different IP. The table can be re-initialized with different configuration.

Problem solved here is that the table info is reset to no configuration before re-initializing. E.g. the snmp indexes are added during initialization, but not removed in shutdown. To make sure we clear all configuration from the table, a memset() was chosen: we start with a clean table.


bvanassche:
Thanks for the patch!

PR #785, opened 2024-02-03
Add Linux 6.7 compatibility parsing /proc/net/snmp

F-i-f:
Linux 6.7 adds a new OutTransmits field to Ip in /proc/net/snmp. This breaks the hard-coded assumptions about the Ip line length. Add compatibility to parse Linux 6.7 Ip header while keep support for previous versions.

### DIFF
--- a/agent/mibgroup/ip-mib/data_access/systemstats_linux.c
+++ b/agent/mibgroup/ip-mib/data_access/systemstats_linux.c
@@ -36,7 +36,7 @@ netsnmp_access_systemstats_arch_init(void)
 }
 
 /*
-  /proc/net/snmp
+  /proc/net/snmp - Linux 6.6 and lower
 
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

ColdForest:
This actually appeared in the v6.6.x kernel series (comments should reference this version instead of v6.7.x). [^fn1]

We see these same errors occurring in systems running our v6.6.x kernel.

[^fn1]: [net: fix IPSTATS_MIB_OUTPKGS increment in OutForwDatagrams.](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/net/ipv4/proc.c?h=linux-6.6.y&id=56712f74b7046b36d142f2f40613566afa0618df)

bvanassche:
@ColdForest , please check whether the Net-SNMP master branch works fine with kernel v6.6.x.

Issue #700, opened 2023-09-18
Memory leak when parsing arguments in snmplib/snmp_api.c:7963


moshekaplan:
Tested on 13622d11459454bf435bc8afeaacb6a740ddf6cf

```
$ '/home/user/Desktop/net-snmp_asan/apps/snmpget' '' '-xd' 'A' '-xd' 'B'
MIB search path: /home/user/.snmp/mibs:/usr/local/share/snmp/mibs
Cannot find module (SNMPv2-MIB): At line 0 in (none)
Cannot find module (IF-MIB): At line 0 in (none)
Cannot find module (IP-MIB): At line 0 in (none)
Cannot find module (TCP-MIB): At line 0 in (none)
Cannot find module (UDP-MIB): At line 0 in (none)
Cannot find module (HOST-RESOURCES-MIB): At line 0 in (none)
Cannot find module (NOTIFICATION-LOG-MIB): At line 0 in (none)
Cannot find module (DISMAN-EVENT-MIB): At line 0 in (none)
Cannot find module (DISMAN-SCHEDULE-MIB): At line 0 in (none)
Cannot find module (UCD-SNMP-MIB): At line 0 in (none)
Cannot find module (UCD-DEMO-MIB): At line 0 in (none)
Cannot find module (SNMP-TARGET-MIB): At line 0 in (none)
Cannot find module (NET-SNMP-AGENT-MIB): At line 0 in (none)
Cannot find module (HOST-RESOURCES-TYPES): At line 0 in (none)
Cannot find module (SNMP-FRAMEWORK-MIB): At line 0 in (none)
Cannot find module (SNMP-MPD-MIB): At line 0 in (none)
Cannot find module (SNMP-USER-BASED-SM-MIB): At line 0 in (none)
Cannot find module (SNMP-VIEW-BASED-ACM-MIB): At line 0 in (none)
Cannot find module (SNMP-COMMUNITY-MIB): At line 0 in (none)
Cannot find module (IPV6-ICMP-MIB): At line 0 in (none)
Cannot find module (IPV6-MIB): At line 0 in (none)
Cannot find module (IPV6-TCP-MIB): At line 0 in (none)
Cannot find module (IPV6-UDP-MIB): At line 0 in (none)
Cannot find module (IP-FORWARD-MIB): At line 0 in (none)
Cannot find module (NET-SNMP-PASS-MIB): At line 0 in (none)
[...]

PR #791, opened 2024-02-26
Fix leaks in snmpv3.c

ashamedbit:
Hi,
I am a new contributor looking forward to contribute to the shc project :).

This commit fixes leak described in #700.
$ './net-snmp/apps/snmpget' '' '-xd' 'A' '-xd' 'B'
is enough to trigger the leak.

In the case of 'x' option the leak occurs because `session->securityPrivProto` is not freed before assigning. The patch corrects this.
It also fixes another similar leak in case of 'a' argument. Similarly in this case `session->securityAuthProto` is not free before assigning,

Let me know if this is a good patch! If not let me know what changes need be done. I am eager to contribute more to this repository :)


bvanassche:
Thanks for the patch! I merged this patch because it seems to me that all relevant assignments assign dynamically allocated memory.

PR #794, opened 2024-02-29
libsnmp: Reassign timeout value for select(2)

spellingmistake:
As select may update the timeout value, the code may not sleep for 100 ms. If timeout is updated, the code will only sleep for 1 ms once and the immediately return from select(2).
Reassign the timeout value in the loop shouldn't be harmful even on systems that don't update the timeout value.

### DIFF
--- a/snmplib/callback.c
+++ b/snmplib/callback.c
@@ -118,7 +118,7 @@ NETSNMP_STATIC_INLINE int
 _callback_lock(int major, int minor, const char* warn, int do_assert)
 {
     int lock_holded=0;
-    NETSNMP_SELECT_TIMEVAL lock_time = { 0, 1000 };
+    NETSNMP_SELECT_TIMEVAL lock_time;
 
 #ifdef NETSNMP_PARANOID_LEVEL_HIGH
     if (major >= MAX_CALLBACK_IDS || minor >= MAX_CALLBACK_SUBIDS) {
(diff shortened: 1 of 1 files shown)


snmplib/callback.c
```diff
@@ -132,8 +132,10 @@ _callback_lock(int major, int minor, const char* warn, int do_assert)
                 types[major], (SNMP_CALLBACK_LIBRARY == major) ?
                 SNMP_STRORNULL(lib[minor]) : "null"));
 #endif
-    while (CALLBACK_LOCK_COUNT(major,minor) >= 1 && ++lock_holded < 100)
+    while (CALLBACK_LOCK_COUNT(major,minor) >= 1 && ++lock_holded < 100) {
+	lock_time = (NETSNMP_SELECT_TIMEVAL) { 0, 1000 };
```
bvanassche, line 136: Please stick to the ANSI C standard. This type of cast is not ANSI C code.

spellingmistake:
OK, no problem I went for the approach to indivdually assign the members, assuming that `struct __ms_timeval` has the same member names. I didn't put the declaration and initialization in the while block because it looks like the rest of the code does declarations only at the beginning of functions.
Sorry for the multiple forced pushes. Took me a while to fix my broken gpg-key setup.

bvanassche:
Thanks for the patch!

PR #796, opened 2024-03-03
snmpd/snmptrapd.service: Remove dead syslog.target

C0rn3j:
https://github.com/systemd/systemd/blob/6aa8d43ade72e24c9426e604f7fc4b7582b9db7c/NEWS#L72-L73
this target hasn't existed for over a decade


bvanassche:
Thanks for the patch!

PR #801, opened 2024-03-19
V5-9-patches backports from master

nielsb:


bvanassche:
Thanks for having backported these patches.

PR #805, opened 2024-04-08
A set of clean-up commits for V5-9-patches

nielsb:
Mostly trivial fixes :-)


bvanassche:
Thanks for the patches!

PR #824, opened 2024-06-11
Better test and info for libnl-3 presence

nielsb:


agent/mibgroup/etherlike-mib/data_access/dot3stats_linux.c
```diff
@@ -6,9 +6,19 @@
...
+#if !(defined(HAVE_LIBNL3) && defined(HAVE_NETLINK_ROUTE_LINK_H))
+ #ifndef HAVE_LIBNL3
+  #error libnl3-devel (RedHat derivatives) / libnl-3-dev (Debian derivatives) required
+ #endif
+ #ifndef HAVE_NETLINK_ROUTE_LINK_H
+  #error libnl-route-3-dev required (Debian derivatives)
+ #endif
+ #if !defined(HAVE_LIBNL3) && !defined(HAVE_NETLINK_ROUTE_LINK_H)
+  #error pkg-config/pkgconfig is required to detect libnl-3
+ #endif
```
bvanassche, line 19: Please add ` && defined(__linux__)`. libnl is supported on Android but pkg-config not. It is only on Linux that pkg-config is required to detect libnl / libnl-route.

PR #825, opened 2024-06-13
Better test and info for libnl-3 presence

nielsb:


bvanassche:
Thanks for the patch!

PR #834, opened 2024-07-30
net-snmp-create-v3-user has an option to use -x AES128, but it was not reflected in usage hint

ivan-antonyuck:
net-snmp-create-v3-user has an ability to use AES128, but does not advertise it


hardaker:
Thanks for catching that!

PR #835, opened 2024-07-31
FreeBSD: Fix the sysctl() configure test

nielsb:
Forward a patch from 2020 that fixed the sysctl for FreeBSD. This allows compilation to complete for current FreeBSD releases.


bvanassche:
Thanks for the patch!

PR #838, opened 2024-08-09
Update man page for net-snmp-create-v3-user

azertyfun:
Hi! Noticed the manpage was missing some options for `net-snmp-create-v3-user`:

```
Usage:
  net-snmp-create-v3-user [-ro] [-A authpass] [-X privpass]
                          [-a MD5|SHA|SHA-512|SHA-384|SHA-256|SHA-224] [-x DES|AES|AES128] [username]
```

Missing option for authentication algorithm and missing parameters for encryption algorithms have been documented.

Although case does not matter for `-A` and `-X` (which seems to be its own can of worms that needs to be kept for compatibility), I've switched them around to match the usage string of `net-snmp-create-v3-user --help`.


hardaker:
Patch looks good thank you!

PR #843, opened 2024-08-22
Fix ipv6.c for DragonflyBSD (#841)

nielsb:
Fixes: ebb758e3 (x Fix var_udp6 and var_tcp6 on FreeBSD)

### DIFF
--- a/agent/mibgroup/mibII/ipv6.c
+++ b/agent/mibgroup/mibII/ipv6.c
@@ -1516,7 +1516,7 @@ var_udp6(register struct variable * vp,
     int             result;
     int             i, j;
     caddr_t         p;
-#if defined(openbsd4)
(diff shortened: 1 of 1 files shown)


bvanassche:
Although your work is highly appreciated, how popular is DragonflyBSD these days and how many DragonflyBSD users use Net-SNMP? I'm wondering whether we should keep supporting Net-SNMP on DragonflyBSD.

nielsb:
Itis there alright, but it has been a while since the last distribution.
https://distrowatch.com/table.php?distribution=dragonflybsd
I set it up a while back in my test, and I dont like that it was just killed by mistake.
I have no idea about the popularity, I just routinely test FreeBSD, NetBSD, OpenBSD, DragonFlyBSD, Fedora, Debian, Ubuntu, Alma and Raspian :-)

PR #859, opened 2024-10-04
notificationFilter: Fix mask data copying

kam3nskii:
Hello,
I'm a new contributor looking forward to correct some behavior I found.
 
After using the `notificationFilter` directive for creating SNMP-NOTIFICATION-MIB notification filter profile, the `snmpNotifyFilterTable` always showed entries with an empty mask (`0x`). The problem was an uninitialized length field.


bvanassche:
Thanks for the patch!

PR #860, opened 2024-10-07
Port sshtosnmp to FreeBSD.

nielsb:
Completes 3346619 (libsnmp: Port the SSH domain transport to FreeBSD)

### DIFF
--- a/agent/mibgroup/mibII/ipv6.c
+++ b/agent/mibgroup/mibII/ipv6.c
@@ -1516,7 +1516,7 @@ var_udp6(register struct variable * vp,
     int             result;
     int             i, j;
     caddr_t         p;
-#if defined(openbsd4)
+#if defined(openbsd4) || defined(dragonfly)
     static struct inpcb in6pcb, savpcb;
 #elif defined(freebsd3)
     static struct xinpcb in6pcb, savpcb;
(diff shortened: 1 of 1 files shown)


apps/sshtosnmp.c
```diff
@@ -124,21 +137,35 @@ main(int argc, char **argv) {
     {
         struct msghdr msg = { 0 };
         struct iovec iov = { buf, buf_len };
-        struct ucred ouruser = { 0 };
         struct cmsghdr *cmsg;
-	char cmsgbuf[CMSG_SPACE(sizeof(ouruser))];
```
bvanassche, line 129: I'd like to preserve the `CMSG_SPACE(sizeof(ouruser))` expression for the Linux case. Please move the `cmsgbuf` declaration into the `#ifdef` statement such that it can be preserved.

apps/sshtosnmp.c
```diff
@@ -124,21 +137,35 @@ main(int argc, char **argv) {
...
+	char cmsgbuf[2*4096];
+#if defined(SCM_CREDENTIALS)
+	struct ucred ouruser = { 0 };
```
bvanassche, line 143: If we know that code is built for Linux, `{ 0 }` can be simplified into `{}`. Although this is a C23 feature, Linux C compilers support this feature since considerable time.

nielsb: But do we know that? Others might implement SCM_CREDENTIALS?

bvanassche: That's fair. Let's keep `{ 0 }`.

PR #876, opened 2024-10-28
Update extend.c

DmitriyFedin:
Potential DEREF_OF_NULL if strdup() fails to allocate memory, it will return NULL. The patch added a check for the return value.

Found by Linux Verification Center (linuxtesting.org) with SVACE.
Reporter: Dmitrii Fedin ([d.fedin@fobos-nt.ru](mailto:d.fedin@fobos-nt.ru)).


ProjectMutilation:
@DmitriyFedin please rewrite the commit message - something like "agent: avoid potential null derefence"

ProjectMutilation (approved):

hardaker:
Thanks!

PR #891, opened 2024-11-10
Fix memory leak of tclist

cmb69:
`tclist` is dynamically allocated in `netsnmp_init_mib_internals()`, which is called unconditionally from `netsnmp_init_mib()`.  As such the memory should be freed in `shutdown_mib()`.

### DIFF
--- a/include/net-snmp/library/parse.h
+++ b/include/net-snmp/library/parse.h
@@ -45,6 +45,8 @@ SOFTWARE.
 #define MAXLABEL        NETSNMP_MAXLABEL
 #endif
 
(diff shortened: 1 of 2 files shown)


bvanassche:
Why has a new `SNMP_FREE(tclist)` call been added instead of calling `unload_all_mibs()`?

cmb69:
Because I've noticed this with Net-SNMP 5.9.4, and missed that the issue already has been properly fixed via 4bd0d9a8a2860c2c46307aef5ee1ccc69f7e3b62. Sorry for the noise!

PR #909, opened 2024-12-17
Add gitattributes file to use eol=lf for almost all files

orgads:
This fixes autoreconf on MSYS2 and possibly other issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+* eol=lf
+*.bat eol=crlf
+/win32/dist/README* eol=crlf
+*.gif -text
+*.jpg -text
+*.bin -text
+*.bmp -text


bvanassche:
eol=lf applies to all Net-SNMP source files except the unmaintained win32/.../*.ds[pw] files. Shouldn't eol=lf be applied to all Net-SNMP source files?

orgads:
I tried to keep the attributes for minimal. Source files don't have to be lf. If you prefer, I can change it to *, and exclude ds[pw] files.

orgads:
> eol=lf applies to all Net-SNMP source files except the unmaintained win32/.../*.ds[pw] files. Shouldn't eol=lf be applied to all Net-SNMP source files?

Done

bvanassche:
Thanks for the patch!

PR #910, opened 2024-12-17
configure: Fix default paths on mingw32 build

orgads:
The default prefix is typically /mingw32 or /mingw64, which cannot be converted in runtime. Do the conversion in configure.

For example, this change fixes this:
```c
  #define NETSNMP_DEFAULT_MIBDIRS "$HOME/.snmp/mibs;/mingw32/share/snmp/mibs"
```

to be:
```c
  #define NETSNMP_DEFAULT_MIBDIRS "$HOME/.snmp/mibs;C:/msys64/mingw32/share/snmp/mibs"
```

### DIFF
--- a/configure
+++ b/configure
@@ -17776,6 +17776,9 @@ while test "x$tmpset" != "x$SNMPLIBPATH"; do
     SNMPLIBPATH="$tmpset"
     eval tmpset="$tmpset"
 done
+if test "x$PARTIALTARGETOS" = "xmingw32" -o "x$PARTIALTARGETOS" = "xmingw32msvc"; then
+    SNMPLIBPATH="$(cygpath --mixed "$SNMPLIBPATH")"
+fi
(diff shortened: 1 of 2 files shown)


orgads:
This is a partial solution. It only works if MSYS2 is installed on the target machine in the same location as the build machine. I wonder if there's a way to resolve this in runtime.

orgads:
The best way I can think of is using paths that are relative to the binary, e.g. `../share/snmp/mibs`, then resolve them in runtime. But this is not that simple, because the user can specify relative paths in command-line, and in this case, the lookup is expected to be relative to the cwd.

@bvanassche or anybody else, if you have other suggestions, please share them :)

bvanassche:
Is it possible to derive the /mingw32 / /mingw64 prefix from an environment variable?

orgads:
Not on runtime. The binary is win32 native. 

Maybe I'll use relative path, and seek relative paths both from cwd and from bin?

bvanassche:
A plain MinGW installation does not include `cygpath`, isn't it?

PR #911, opened 2024-12-17
libsnmp: Support $HOME in MIBDIRS also on Windows

orgads:


bvanassche:
Thanks for the patch!

PR #912, opened 2024-12-18
libsnmp: Fix homedir lookup also for configuration

orgads:


apps/encode_keychange.c
```diff
@@ -103,6 +103,15 @@ char           *snmp_getpassphrase(const char *prompt, int fvisible);
 static char    *getpass(const char *prompt);
 #endif
 
+static char *gethomedir() {
```
bvanassche, line 106: Instead of duplicating the netsnmp_gethomedir() implementation, please call netsnmp_gethomedir(). apps link with libsnmp anyway.

include/net-snmp/library/tools.h
```diff
@@ -290,6 +290,8 @@ extern          "C" {
     NETSNMP_IMPORT
     int             netsnmp_string_time_to_secs(const char *time_string);
 
+    char            *netsnmp_gethomedir();
```
bvanassche, line 293: NETSNMP_IMPORT should be added to make this function available to apps when building libsnmp as a Windows DLL.

snmplib/tools.c
```diff
@@ -1413,3 +1413,12 @@ netsnmp_string_time_to_secs(const char *time_string) {
     return secs;
 }
 #endif /* NETSNMP_FEATURE_REMOVE_STRING_TIME_TO_SECS */
+
+char *netsnmp_gethomedir() {
```
bvanassche, line 1417: Why the return type `char *` instead of `const char *`? Callers of this function are not expected to modify the string returned by this function, isn't it?

bvanassche (changes requested):

include/net-snmp/library/tools.h
orgads: It wasn't used outside the lib. Now it is, added.

apps/encode_keychange.c
orgads: Done

snmplib/tools.c
orgads: Done

bvanassche:
Thanks for the patch!

PR #923, opened 2025-01-18
Add option to exclude Network Block Devices

nuxi:
On Linux, loading the `nbd` driver creates a number of `nbd` devices even if they aren't actually in use. On top of that, QEMU supports using them as fancier loopback devices for manipulating QCOW2 images. So even when they are in use, there may be reasons to exclude them from the SNMP data.

Since I was updating the manpage, I also added the missing documentation for `diskio_exclude_md` from PR #511 

I'm also not a troff expert in any way (only reason I could update the manpage was that it was a copy and paste job) but ending a line in a period seems to insert an extra space into the text. So I rewrapped the text so that none of these lines end in `e.g.`


bvanassche:
Thanks for the patches!

PR #924, opened 2025-01-19
snmplib: Fix potential memory leak

maks-mishin:
Dynamic memory, referenced by 'dup->container_name', is allocated by calling function 'netsnmp_container_data_dup' and lost at container_binary_array.c:794.


snmplib/container_binary_array.c
```diff
@@ -789,6 +789,7 @@ _ba_duplicate(netsnmp_container *c, void *ctx, u_int flags)
     dupt->data = (void**) malloc(dupt->max_size * sizeof(void*));
     if (NULL == dupt->data) {
         snmp_log(LOG_ERR, "no memory for binary array duplicate\n");
+        free(dup->container_name);
```
bvanassche, line 792: Why has this free() call been inserted before the netsnmp_binary_array_release() call instead of inside the netsnmp_binary_array_release() call?

PR #925, opened 2025-01-22
Fix include stdint.h missing to get SIZE_MAX

zodf0055980:
This fixes commit d33578d for ubuntu building.
This problem is same as #839


hardaker:
Thanks!

PR #928, opened 2025-01-31
Avoid unused variable warning

egilkv:
Happens on latest Rpi compiler. Also trivial update of INSTALL


bvanassche:
Thanks for the patches!

PR #929, opened 2025-02-03
agent: mibgroup: mibII: fix prevent use-after-free in write_vacmAccessStatus

AntonMoryakov:
1. Added error checking in the `ACTION`, `COMMIT`, and `UNDO` blocks. If `access_parse_oid` returns an error, the function terminates with `SNMP_ERR_INCONSISTENTNAME`.

2. Memory allocated for `newGroupName` and `newContextPrefix` is freed only after it is no longer needed (e.g., after calling `vacm_getAccessEntry`).

3. The pointers `newGroupName` and `newContextPrefix` are no longer used after calling `free`.


bvanassche:
Thanks for the patch!

PR #938, opened 2025-02-20
mib2c-update: fix ls breakage introduced in a8cbc8db2e971990ad

funman:
$UPDATE_CURR is an absolute path, prepending ./ does not make sense


bvanassche:
Thanks for the patch!

PR #941, opened 2025-03-05
snmplib: add deallocation of container name

ProjectMutilation:
The function netsnmp_binary_array_release() does not deallocate dynamic char array.
Closes https://github.com/net-snmp/net-snmp/pull/924


bvanassche:
Thanks for the patch!

PR #946, opened 2025-03-08
support openbsd8

sthen:
Preemptive, but OpenBSD follows a roughly 6 month release schedule with a simple 0.1 increment each time and there are no plans to diverge from this, so this will be needed in the future. (No big hurry, development code just moved to using 7.7 for a release in a month or so, but seemed worth including as you're working on a release).


hardaker:
seems wise to get this in early, thanks.

PR #947, opened 2025-03-08
use correct filename when installing generic-get-in_addr_t.m2i

sthen:
d3a541a added local/mib2c-conf.d/generic-get-in_addr_t.m2i but the Makefile tries to install generic-get-inaddr_t.m2i resulting in "install: ./mib2c-conf.d/generic-get-inaddr_t.m2i: No such file or directory"


bvanassche:
Thanks for the patch!

PR #948, opened 2025-03-14
Another slew of backports from master to V5-9-patches

nielsb:


bvanassche:
Thanks for the backports!

PR #953, opened 2025-03-28
Better testing for syslog.h presence

klausholstjacobsen:
Added better testing for <sys/syslog.h> or <syslog.h>


bvanassche:
Thanks for the patch!

PR #959, opened 2025-04-11
snmplib: scapi.c: fix unchecked function result

AntonMoryakov:
Checking the return values ​​of MD5_Update and SHA1_Update in the
sc_hash_type() function. In case of an error, returning SNMPERR_GENERR.

Previously, calls to these functions were not checked, which could lead to
undefined behavior in case of hash errors.

Signed-off-by: Anton Moryakov [ant.v.moryakov@gmail.com](mailto:ant.v.moryakov@gmail.com)

### DIFF
--- a/snmplib/scapi.c
+++ b/snmplib/scapi.c
@@ -1001,7 +1001,9 @@ sc_hash_type_quit:
         if (*MAC_len < MD5_DIGEST_LENGTH)
             return (SNMPERR_GENERR);      /* the buffer isn't big enough */
         MD5_Init(&cmd5);
-        MD5_Update(&cmd5, buf, buf_len);
+        ret = !MD5_Update(&cmd5, buf, buf_len);
+        if (ret != 0)
+            return SNMPERR_GENERR;
         MD5_Final(MAC, &cmd5);
         *MAC_len = MD5_DIGEST_LENGTH;
     } else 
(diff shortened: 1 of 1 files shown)


bvanassche:
Are you sure that this patch is needed? In the openssl library I see that this function always returns 1. See also https://github.com/openssl/openssl/blob/8d2e4d6d8c927f05948e048fcbf62982feaf11b4/include/crypto/md32_common.h#L154.

snmplib/scapi.c
```diff
@@ -1001,7 +1001,9 @@ sc_hash_type(int auth_type, const u_char * buf, size_t buf_len, u_char * MAC,
         if (*MAC_len < MD5_DIGEST_LENGTH)
             return (SNMPERR_GENERR);      /* the buffer isn't big enough */
         MD5_Init(&cmd5);
-        MD5_Update(&cmd5, buf, buf_len);
+        ret = !MD5_Update(&cmd5, buf, buf_len);
+        if (ret != 0)
+            return (SNMPERR_GENERR);
```
bvanassche, line 1006: Here and below: `return` is not a function name. Please remove the superfluous parentheses.

AntonMoryakov: Removed superfluous brackets

bvanassche:
Thanks for the patch!

PR #962, opened 2025-04-12
Fix memory corruptio in snmp_sess_read2 and _sess_read_dgram_packet

feroz-ahmed:


bvanassche:
Please combine the three commits into a single commit, e.g. with `git rebase -i HEAD~3`.

snmplib/snmp_api.c
```diff
@@ -1926,7 +1926,7 @@ snmp_free_session(netsnmp_session * s)
          */
         netsnmp_callback_clear_client_arg(s, 0, 0);
 
-        free((char *) s);
```
bvanassche, line 1929: `free(NULL)` has no effect and clearing a variable that won't be reused is not necessary. Hence, this change of `free()` into `SNMP_FREE()` is not useful. Same comment for the other `SNMP_FREE()` change.

snmplib/snmp_api.c
```diff
@@ -6064,8 +6064,10 @@ _sess_read_dgram_packet(void *sessp, netsnmp_large_fd_set * fdset,
                                               SNMP_MAX_RCV_MSG_SIZE,
                                               &rcvp->opaque, &rcvp->olength);
     if (rcvp->packet_len == -1) {
-        sp->s_snmp_errno = SNMPERR_BAD_RECVFROM;
```
bvanassche, line 6067: Please include a root-cause analysis and/or reproducer. I'm not aware of any code in the snmplib/transports directory that closes a session. Did I perhaps overlook something?

bvanassche (changes requested):

snmplib/snmp_api.c
```diff
@@ -6064,8 +6064,10 @@ _sess_read_dgram_packet(void *sessp, netsnmp_large_fd_set * fdset,
...
-        sp->s_errno = errno;
+        if (!sp) {
```
ProjectMutilation, line 6067: It looks like an redundant check. If sp is NULL, then the execution is terminated early.

PR #966, opened 2025-05-01
Fixed an issue in mib2c-update with broken search path

kmeinhar:
The mib2c tool generates source/header files that do not include the name of the MIB (Before `::`).
With this patch it is now possible to specify the generated file name. By default the MIB name is just stripped away.


kmeinhar:
Please let me know if there are any other locations where it might be necessary to do changes like documentation.

hardaker:
Thanks!

Issue #955, opened 2025-04-01
SEGFAULT in SNMP.so (__snmp_xs_cb)


blaaaaab:
Hello,

I was upgrading my system from an old version of net-snmp and with newer version my script segfaults.
With git bisect, I found that the commit 581698a162b47b08f8571c7e1b8991df0764e5d3 seems to introduce the issue.

Here is the stack trace :
0x00007f49cc9d2b9c __snmp_xs_cb (SNMP.so + 0xab9c)
0x00007f49cc572419 snmp_sess_close (libnetsnmp.so.40 + 0x39419)
0x00007f49cc9dedc0 XS_SnmpSessionPtr_DESTROY (SNMP.so + 0x16dc0)
0x00007f49cc708437 Perl_pp_entersub (libperl.so.5.32 + 0x108437)
0x00007f49cc67b18a Perl_call_sv (libperl.so.5.32 + 0x7b18a)
0x00007f49cc710920 S_curse.lto_priv.0 (libperl.so.5.32 + 0x110920)
0x00007f49cc711068 Perl_sv_clear (libperl.so.5.32 + 0x111068)
0x00007f49cc70ff52 Perl_sv_free2 (libperl.so.5.32 + 0x10ff52)
0x00007f49cc73a66f Perl_free_tmps (libperl.so.5.32 + 0x13a66f)
0x00007f49cc9dfe97 XS_SNMP__main_loop (SNMP.so + 0x17e97)
0x00007f49cc708437 Perl_pp_entersub (libperl.so.5.32 + 0x108437)
0x00007f49cc7001a0 Perl_runops_standard (libperl.so.5.32 + 0x1001a0)
0x00007f49cc67e90c perl_run (libperl.so.5.32 + 0x7e90c)
0x0000559bc099e34a main (perl + 0x134a)
0x00007f49cc2295d0 __libc_start_call_main (libc.so.6 + 0x295d0)
0x00007f49cc229680 __libc_start_main@@GLIBC_2.34 (libc.so.6 + 0x29680)
0x0000559bc099e385 _start (perl + 0x1385)

bvanassche:
Thank you for having bisected this issue.

Please help with testing the latest version of the master branch.

Please also provide a reproducer for this crash.

blaaaaab:
Hello,

The issue is fixed in the latest version of the master branch.
I am unable to provide a reproducer for the crash at the moment, it is part of a bigger application with many dependencies.

[...]

PR #967, opened 2025-05-02
Perl: Do not crash on resend callbacks.

sesse:
Do not act on NETSNMP_CALLBACK_OP_RESEND messages, as they are not final (the callback assumes that we can free the callback struct and generally mutate it freely without ever seeing it again, causing double-frees on resends).

Possibly related to #955.


sesse:
I noticed a patch related to this was never included upstream (I'm fairly certain we sent it to SourceForge, but I guess it was never really picked up), so I'm including it in the PR. It looks big and scary, but is much less so when ignoring indent changes. :-)

sesse:
It seems the first patch was already done for other parts of the code in 2ed4fd6157075676624aa83ce011cdf7dbf8439c.

kajtzu:
Thanks for submitting the patch. I concur, this fixes some issues^Wcrashes that we noticed on EL9 using EL9 net-snmp-5.9.1 on our home-made collector. ;)

bvanassche:
Slightly modified versions of these patches have been applied on the master branch. Please help with testing the master branch.

sesse:
I tried compiling master, but getting it to work on my system was basically hours of pain, so sorry. :-)

PR #969, opened 2025-05-15
pingCtlTable.c: Avoid memory leak if strdup fails

moshekaplan:
Avoid memory leak if strdup fails.
Fixes Coverity 454615


hardaker:
Also looks good but we should make these SNMP_FREEs

PR #970, opened 2025-05-15
pingCtlTable.c: Avoid memory leak if an allocation fails

moshekaplan:
Fixes Coverity 454608

### DIFF
--- a/agent/mibgroup/disman/ping/pingCtlTable.c
+++ b/agent/mibgroup/disman/ping/pingCtlTable.c
@@ -241,6 +241,7 @@ create_pingCtlTable_data(void)
         StorageNew->pingCtlDescr == NULL ||
         StorageNew->pingCtlSourceAddress == NULL) {
         free(StorageNew->pingCtlTargetAddress);
+        free(StorageNew->pingCtlDataFill);
         free(StorageNew->pingCtlTrapGeneration);
         free(StorageNew->pingCtlType);
         free(StorageNew->pingCtlDescr);


hardaker:
Merging -- though these should really be SNMP_FREEs as if the above malloc fails we'll now just crash here instead if the previous strdup returned a NULL.

moshekaplan:
@hardaker : I apologize for my ignorance here, but if `malloc` fails, wouldn't it return `NULL`, and if you call `free(NULL)`, it's a no-op?

I'm basing this on this man page (https://man7.org/linux/man-pages/man3/free.3.html) which states:
> The free() function frees the memory space pointed to by ptr,
       which must have been returned by a previous call to malloc() or
       related functions.  Otherwise, or if ptr has already been freed,
       undefined behavior occurs.  **If ptr is NULL, no operation is
       performed.**
       
and:
> The malloc(), calloc(), realloc(), and reallocarray() functions
       return a pointer to the allocated memory, which is suitably
       aligned for any type that fits into the requested size or less.
       **On error, these functions return NULL** and set [errno](https://man7.org/linux/man-pages/man3/errno.3.html).

bvanassche:
free(NULL) is not only supported on Linux. It is required by POSIX that free(NULL) is a no-op. See also https://pubs.opengroup.org/onlinepubs/007908799/xsh/free.html.

hardaker:
@bvanassche is correct and is actually the whole reason we originally created the SNMP_FREE macro:

```
#define SNMP_FREE(s)    do { if (s) { free(s); s=NULL; } } while(0)
```

Which is basically a wrapper around free() that checks whether or not the pointer is already NULL (and then sets to NULL after that).

There are a number of OSes that likely check if you're calling `free(0)` or not, but there are certainly a large number that don't check as well.  We try to support them all (which makes things difficult at times, certainly).

bvanassche:
From a high-quality LLM (LLMs may make mistakes):
```
Which C compilers and/or operating systems did not support "free(NULL)"?

The C standard, specifically since ANSI C (C89/C90) and later versions like C99 and C11, explicitly states that free(NULL)
should perform no action. This means that a compliant C compiler and its associated library will simply ignore a call to
free with a null pointer, without causing any errors or crashes.

However, historical and non-compliant C implementations did exist where calling free(NULL) could lead to crashes or
undefined behavior. While it's difficult to compile an exhaustive list, here's what we know:

Older C Libraries (Pre-ANSI C): Some very old C libraries, particularly from the 1980s or earlier (e.g., implementations
based on the original K&R C book), did not have the explicit free(NULL) no-op guarantee. In such environments,
attempting to free(NULL) could indeed crash the program.

Specific Embedded Systems/Compilers: In highly constrained or specialized embedded systems, or with compilers
targeting such platforms, memory management might be very custom and not strictly adhere to later C standards. It's
conceivable that some of these non-standard environments might have had issues with free(NULL).

Non-standard or Buggy Implementations: Any compiler or library that didn't fully conform to the C standard (even after
the standard was established) could potentially exhibit this problematic behavior. This would be a bug in that specific
implementation rather than a general characteristic of a compiler or operating system.

In summary:

For any modern, standard-compliant C compiler and operating system (including GCC, Clang, Visual C++, and various
Unix-like systems, Windows, macOS, and modern embedded OSes), free(NULL) is guaranteed to be safe and do nothing.
The primary reason for this behavior in the standard is to simplify cleanup code, allowing programmers to call free on
pointers that might or might not have been successfully allocated, without needing an explicit if (ptr != NULL) check.

If you encounter a system where free(NULL) causes a crash, it's almost certainly an outdated, non-compliant, or buggy
C library implementation.
```

If someone would like to do this, I wouldn't object against dropping support for C libraries that do not support `free(NULL)`.

hardaker:
I would object.  Net-SNMP is used in a lot of embedded OSes that are potentially using outdated compilers.

PR #971, opened 2025-05-15
override.c: Avoid leaking memory if an allocation fails

moshekaplan:
Fixes Coverity 454607

### DIFF
--- a/agent/mibgroup/utilities/override.c
+++ b/agent/mibgroup/utilities/override.c
@@ -240,6 +240,8 @@ netsnmp_parse_override(const char *token, char *line)
     the_reg = SNMP_MALLOC_TYPEDEF(netsnmp_handler_registration);
     if (!the_reg) {
         config_perror("memory allocation failure");
+        if (thedata->type != ASN_NULL)
+            free(thedata->value);
         free(thedata);
         return;
     }
@@ -259,6 +261,8 @@ netsnmp_parse_override(const char *token, char *line)
         SNMP_FREE(the_reg->handlerName);
         SNMP_FREE(the_reg);
         config_perror("memory allocation failure");
+        if (thedata->type != ASN_NULL)
+            free(thedata->value);
         free(thedata);
         return;
     }


hardaker:
Took a very quick look, which wast long enough to think this may not work as intended -- value isn't always allocated, and thus sholudn't always be freed.

moshekaplan:
> Took a very quick look, which wast long enough to think this may not work as intended -- value isn't always allocated, and thus sholudn't always be freed.

AFAICT, the only time `thedata->value` is not allocated is if the `type` is `ASN_NULL`. However, `thedata` is allocated with `SNMP_MALLOC_TYPEDEF` which calls `calloc` and sets the allocated memory to zero. As such, `thedata->value` has a default value of `0`. My assumption here is that `NULL == 0`. Assuming that is correct, in the event that the `type` is `ASN_NULL`. we'd be calling `free(NULL)`, which should be a no-op.

The one thing is that if we're going to be concerned that `NULL` equals something besides 0, we might want to redefine SNMP_FREE. Currently it is the following:
```#define SNMP_FREE(s)    do { if (s) { free(s); s=NULL; } } while(0)```
If that is a concern, it might be worth changing it to:
```#define SNMP_FREE(s)    do { if (s != NULL) { free(s); s=NULL; } } while(0)```

moshekaplan:
@hardaker : Sorry for the long turnaround on this. I've added a check to see if the type is `ASN_NULL` . Could you please recheck this to see if it operates correctly?

bvanassche:
Please combine the three commits into a single commit to make it easier to review these.

moshekaplan:
> Please combine the three commits into a single commit to make it easier to review these.

Done!

bvanassche:
Thanks for the patch!

bvanassche:
Please help with reviewing the following two commits that modify the same function:
* 3ba73ed359661daac7b3da279b9b4cb1313a6e23.
* e0755b54196ccca061c930053701cd2d30d4ab02.

moshekaplan:
> Please help with reviewing the following two commits that modify the same function:
> 
> * [3ba73ed](https://github.com/net-snmp/net-snmp/commit/3ba73ed359661daac7b3da279b9b4cb1313a6e23).
> * [e0755b5](https://github.com/net-snmp/net-snmp/commit/e0755b54196ccca061c930053701cd2d30d4ab02).

FWIW, looks fine to me.

PR #972, opened 2025-05-15
override.c: Avoid leaking memory in rootoid if an allocation fails

moshekaplan:
Fixes Coverity 454597


hardaker (approved):
Looks good thanks!

PR #973, opened 2025-05-15
Fix warnings in V5 9 patches

nielsb:
libssh2 v1.2.8 deprecated libssh2_session_startup,
use libssh2_session_handshake when available.

Fix OpenBSD warning about strcpy, sprintf when runnoig tests.


bvanassche:
These patches have been applied manually on the V5-9-patches and master branches. Thanks for the patches!

PR #974, opened 2025-05-15
snmpset: fix memory leak

AbelAnaya:
Small memory leak fixed when allocating a char* with strdup. The memory is now correctly freed before closing.

I know it is very small and that snmpset binary memory is anyway freed after execution but there are teams using snmpset.c code as a guide to implement snmp set operations in their applications with net-snmp library, and this memory leak could be copied.

### DIFF
--- a/apps/snmpset.c
+++ b/apps/snmpset.c
@@ -131,7 +131,8 @@ main(int argc, char *argv[])
 
     SOCK_STARTUP;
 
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

bvanassche:
`-Werror=declaration-after-statement` has been enabled on the V5-9-patches and master branches to make sure that mixed declarations and statements will be noticed for future pull requests.

PR #981, opened 2025-06-03
win32: add build support for Windows on ARM

hyh19962008:
I was able to produce a successful build without trouble on my Windows 11 ARM64 + MSVC 2022.


bvanassche:
Thanks for the patch!

PR #984, opened 2025-06-10
loadave.c: fix out-of-bounds access in loadave_parse_config

AntonMoryakov:
Static analysis reported a potential buffer overflow when accessing maxload[i - 1] for i = 0.

The issue was fixed by:
- Adding a check to ensure i > 0 before accessing maxload[i - 1]
- Using NETSNMP_DEFMAXLOADAVE as a default value for the first element (i == 0)
- Ensuring that the code does not attempt to access memory outside of the maxload array

This change makes the code more robust and avoids undefined behavior.


bvanassche:
Thanks for the patch!

PR #994, opened 2025-07-24
tunnel.c: Avoid leaking tunnel->ifname

moshekaplan:
Fixes coverity 454603


bvanassche:
Thanks for the patch!

PR #1001, opened 2025-08-30
Implement ASN_FLOAT as a possible type for the Perl API setValue

### LINKED ISSUE #1000: setValue in Perl->C agent interface does not implement data type ASN_FLOAT
When trying to do something like

```
$request->setValue(ASN_FLOAT, -6.6);
```
in a Perl snmpd agent, I. do get a `unknown var value type: 72` back.
I think I found the cause, which is the missing implementation of `ASN_FLOAT` in `perl/agent/agent.xs`.

RincewindsHat:
This is a attempt to fix #1000.

I will honestly start with acknowleding, that I do not know much about Perl
and bascially just c&p some other pieces of code together.

Also, I did not test this changes.
Maybe someone can help with that.

### DIFF
--- a/perl/agent/agent.xs
+++ b/perl/agent/agent.xs
@@ -855,7 +855,19 @@ nari_setValue(me, type, value)
 
               RETVAL = 1;
               break;
-              
+			case ASN_FLOAT:
+				/* We want a floating point number here here */
+				if ((SvTYPE(value) == SVt_PVMG) || SvNOK(value)) {
+					dtmp = SvNOK(value);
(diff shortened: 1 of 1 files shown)


hardaker:
So floats really don't exist in SNMP.  Net-SNMP has a hack in place that implements floats via the ASN_OPAQUE_FLOAT type, which is different than ASN_FLOAT.  I think any patch that wants to have the perl agent support for floats should match some of the other examples with how floats are returned in C as well.  See for example the `loadave.c` file and it's use of the opaque float within it.

PR #1003, opened 2025-09-11
build/freebsd: add FreeBSD 16 system header

GenericRikka:
- Add include/net-snmp/system/freebsd16.h (currently inherits freebsd15.h).
- Matches existing pattern for FreeBSD 10–15 headers.

This prepares for FreeBSD 16 while keeping current behavior unchanged.

The patch originated from this FreeBSD PR: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=289351

Signed-off-by: Generic Rikka <rikka.goering@outlook.de>
Tested on: FreeBSD 16-CURRENT


bvanassche:
Thanks for the patch!

PR #1005, opened 2025-09-19
Fix build when USING_AGENTX_MASTER_MODULE is undefined

dyumin:
agentx_parse_agentx_timeout and agentx_parse_agentx_retries defined under USING_AGENTX_MASTER_MODULE


bvanassche:
Thanks for the patch!

PR #1016, opened 2025-10-29
pingCtlTable.c: Fix memory leaks in parse_pingCtlTable()

wxmzy88:
There are memory leaks in the parse_pingCtlTable interface: allocated memory for StorageTmp->pingCtlOwnerIndex, StorageTmp->pingCtlTestName, etc., is not freed in the error paths.


bvanassche:
The incomprehensible comment has been removed and a slightly different patch has been applied. Please take a look at 376682bcab164aa32bc998a6f23e9d85a061f490.

PR #1017, opened 2025-10-30
snmp_agent.c: Fix crash caused by buf being a null pointer

wxmzy88:
If buf = strdup("") failed, then the buf is NULL. When using DEBUGMSGTL to print logs, passing buf as a string parameter when buf is NULL can cause undefined behavior  (typically causing program crash).


bvanassche:
Thanks for the patch!

PR #1020, opened 2025-10-31
diskio_openbsd.c: Add null check after malloc

wxmzy88:
In diskio_getstats(), add null check after malloc dkname, t/tp/te and dk. It can avoid null pointer dereference when alloc memory failed.
Also t should be freed when sysctl(mib, 2, t, &dkn_size, NULL, 0) failed.


bvanassche:
Thanks for the patch!

PR #1022, opened 2025-10-31
libsnmp: Need check env_var before read mib

wxmzy88:
netsnmp_init() must validate env_var is not null before MIB read operations.


bvanassche:
Thanks for the patch!

PR #1023, opened 2025-11-03
Fix null pointer dereference in debug output

AntonMoryakov:
Static analysis reported a possible null pointer dereference in asn_realloc_rbuild_string.

The issue occurs when 'str' is NULL but 'strlength > 0': the pointer is passed to sprint_realloc_asciistring inside a debug block without prior null check, leading to dereference at mib.c:409.

Corrected by adding a null check before calling
sprint_realloc_asciistring, ensuring it is only
invoked when 'str != NULL'.


bvanassche:
Please fix this potential NULL pointer dereference by failing early if `str == NULL && strlength != 0`.

AntonMoryakov:
@bvanassche
I agree with you. The changes are ready.

PR #1024, opened 2025-11-03
fix prevent buffer overflow in hrSWRunParameters assignment

AntonMoryakov:
Static analysis reported a potential buffer overflow in netsnmp_arch_swrun_container_load.

The issue occurs when using snprintf with a
tainted source (process cmdline data) and
improper size handling. Although the format
'%.*s' limits string width, the destination
buffer hrSWRunParameters (size 129) was not
safely bounded by the function call itself,
risking overflow if bounds calculation is
incorrect or bypassed.

Corrected by ensuring safe use of snprintf
with explicit buffer size and proper field
width limit. The call now uses
sizeof(entry->hrSWRunParameters) as the
output bound, while retaining the precision
limit to cap string length at 128 characters,
preventing any out-of-bounds write.

### DIFF
--- a/agent/mibgroup/host/data_access/swrun_procfs_status.c
+++ b/agent/mibgroup/host/data_access/swrun_procfs_status.c
@@ -146,7 +146,8 @@ netsnmp_arch_swrun_container_load( netsnmp_container *container, u_int flags)
                             *cp = ' ';
 
             entry->hrSWRunParameters_len
-                = sprintf(entry->hrSWRunParameters, "%.*s",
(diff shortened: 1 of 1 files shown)


AntonMoryakov:
> The current code can't trigger a buffer overflow. Hence, using the word "fix" in the patch title seems incorrect to me. The proposed change doesn't modify the behavior of the code - all it does is to suppress a complaint from the static analyzer. Has it been considered to improve the static analyzer instead of modifying the Net-SNMP code?

Then you can close this PR. I agree with your opinion.

PR #1025, opened 2025-11-04
fix prevent buffer overflow in ASN_OCTET_STR index allocation

AntonMoryakov:
In register_index(), when generating sequential string indices from 'prev_idx_ptr', the code used a flawed algorithm that could lead to buffer overflow by writing beyond the end of new_index->varbind->buf.

The loop condition relied on buf[i] without proper bounds checking, and the fallback logic for growing the string wrote to buf[i+1] without validating available space.

Rewrite the octet string increment logic to:
- Properly check array bounds using i >= 0
- Use memmove to shift string left when growing
- Validate buffer capacity before extending
- Update val_len correctly

Now returns NULL if buffer is full, preventing memor

### DIFF
--- a/agent/agent_index.c
+++ b/agent/agent_index.c
@@ -346,21 +346,33 @@ register_index(netsnmp_variable_list * varbind, int flags,
             break;
         case ASN_OCTET_STR:
             if (prev_idx_ptr) {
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #1027, opened 2025-11-07
Fix segfaults when `varbind` cannot be constructed

crazyscientist:
For multiple reasons, `varbind` can be a null pointer, which causes segmentation faults.

With this change, the value is checked and the segmentation fault avoided.


bvanassche:
Please run `git pull --rebase` and next force-push this pull request (`git push ${net_snmp_repo_name} +${branch_name}`).

bvanassche:
I was hoping that my commits that I dropped from the master branch would have been removed. Anyway, I cherry-picked your patch on the master branch. Please take a look at commit 03e28b651692419e56ea68642df2bb5b209f2230.

PR #1030, opened 2025-12-02
Fix memory corruption in agentx_got_response

fdumontet6WIND:
we encounter the following crash (onfield)

(gdb) bt
    at ./nptl/pthread_kill.c:44
    at ./nptl/pthread_kill.c:78
    at ./nptl/pthread_kill.c:89
    at ../sysdeps/posix/raise.c:26
    fmt=fmt@entry=0x7e22cf9dbb77 "%s\n") at ../sysdeps/posix/libc_fatal.c:156
    str=str@entry=0x7e22cf9de7b0 "double free or corruption (!prev)")
    at ./malloc/malloc.c:5664
    p=0x56fc941985c0, have_lock=<optimized out>) at ./malloc/malloc.c:4591
    at ./malloc/malloc.c:3391
    dcache=dcache@entry=0x56fc941985d0) at agent_handler.c:934
    session=0x56fc941a5240, reqid=<optimized out>, pdu=<optimized out>,
    magic=0x56fc941985d0) at mibgroup/agentx/master.c:220
--Type <RET> for more, q to quit, c to continue without paging--
    at snmp_api.c:1984
    numfds=0x7fff49e28ee4, fdset=0x7fff49e28f00, timeout=0x7fff49e28ef0,
    block=0x7fff49e28eec, flags=0) at snmp_api.c:6580

in "agentx_got_response" delegated_cache is no more valid and freeing it encounter following error "double free or corruption (!prev)".

memory inspection shows that delegated_cache had been freed and replace by another variable of another type.

thus a copy of that pointer exists and a free occured on it.

Fix: add a refcount in delegated_cache, and increase it when a copy occurs.
     separates the contents from the container by creating a box which has
     a refcount. This ensures that the container is freed when the refcount
     reaches 0.

### DIFF
--- a/agent/mibgroup/agentx/master.c
+++ b/agent/mibgroup/agentx/master.c
@@ -203,6 +203,34 @@ real_init_master(void)
     DEBUGMSGTL(("agentx/master", "initializing...   DONE\n"));
 }
 
+
+static netsnmp_refcnt_void *
+netsnmp_create_delegated_box(void *val_ptr)
+{
+    netsnmp_refcnt_void * box = NULL;
+
+    box = SNMP_MALLOC_TYPEDEF(netsnmp_refcnt_void);
+    if (box) {
+        box->refcnt = 1;
+        box->val = val_ptr;
+    }
+      
+    return box;
+}
+
+static
+netsnmp_delete_delegated_box(netsnmp_refcnt_void * box)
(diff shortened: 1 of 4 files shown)


bvanassche:
Thanks for the patch!

PR #1036, opened 2025-12-25
mibgroup: notification: avoid potential null dereference

ProjectMutilation:
The pointer var_val was utilized before check against NULL.


ProjectMutilation:
@bvanassche hi, can you help with a code review please? I'm not sure that usage of "RS_NONEXISTENT" is correct

ProjectMutilation:
> Is this change needed? My understanding is that `var_val` is never NULL.

I've looked at a lot of possible execution paths and it looks like there really never will be NULL, but I'm not sure.

PR #1042, opened 2026-01-10
Fix zero-width display hint handling per RFC 2579

lukeod:
The current code treats zero-width octet specs (e.g., `0a[`) as malformed and forces width=1. This is incorrect - RFC 2579 Section 3.1 explicitly states "the octet length can be zero."

Zero-width specs are used in practice. For example, RFC 3419 defines:
```
TransportAddressIPv6 ::= TEXTUAL-CONVENTION
    DISPLAY-HINT "0a[2x:2x:2x:2x:2x:2x:2x:2x]0a:2d"
```

The `0a[` and `0a:` emit literal brackets and colons to format addresses as `[2001:db8::1]:80`. With the current code forcing width=1, this produces garbage output.

Changes:
- Remove the incorrect `width = 1` override for zero-width specs
- Handle zero-width by emitting separator/terminator literals without consuming data bytes
- Implement trailing separator/terminator suppression per RFC 2579

Added unit tests covering basic zero-width handling, the RFC 3419 IPv6 format, trailing suppression, repeat indicators, and various edge cases.

### DIFF
--- a/snmplib/mib.c
+++ b/snmplib/mib.c
@@ -512,8 +512,36 @@ sprint_realloc_octet_string(u_char ** buf, size_t * buf_len,
                     term = *hint++;
                 else
                     term = 0;
-                if (width == 0)  /* Handle malformed hint strings */
-                    width = 1;
+            }
+
+            /*
+             * Handle zero-width display hint specs (RFC 2579 Section 3.1).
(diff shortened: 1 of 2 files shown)


bvanassche:
Thanks for the patch!

PR #1048, opened 2026-01-16
Implement force option for net-snmp-create-v3-user

adam-bishop:
In some environments (e.g. Kubernetes) snmpd may be falsely detected as running when the process in question is actually confined within a container.

In these deployments, stopping all snmpd processes is not always feasible as doing so could require terminating containers owned by other cluster users.

This means that I'm unable to create users with the helper script on the container host.

This change introduces a new command-line option (-f) to force user creation, bypassing the check.


bvanassche:
Thanks for the patches!

PR #1049, opened 2026-01-16
agent: Fixed duplicate and invalid free operations in memory allocation failure handling

wxmzy88:


agent/mibgroup/host/hr_disk.c
```diff
@@ -161,8 +161,6 @@ parse_disk_config(const char *token, char *cptr)
                 config_perror("Out of memory");
                 SNMP_FREE(d_new);
                 SNMP_FREE(di_curr);
-                SNMP_FREE(d_set);
-                SNMP_FREE(d_str);
```
bvanassche: `d_set` and `d_str` are initialized to NULL. SNMP_FREE() checks these pointers before calling free(). Even if it wouldn't check these pointers there would be no issue since calling `free(NULL)` is allowed.

bvanassche (changes requested):

agent/mibgroup/host/hr_disk.c
wxmzy88: Although free(NULL) is permitted, it is preferable to remove the redundant free calls for the following two reasons：
（1）Improved Clarity: Removing these statements that never execute makes the error-handling logic more straightforward and easier to understand, avoiding unnecessary confusion for developers.

（2）Enhanced Maintainability: Adhering to the "minimal code" principle allows future memory management audits and code reviews to focus more effectively on genuine resource deallocation issues.

bvanassche:
Thanks for the patch!

PR #1050, opened 2026-01-16
agent/mibgroup/mibII/vacm_conf.c: Fix crash caused by buf being a null pointer

wxmzy88:


bvanassche:
Thanks for the patch!

PR #1053, opened 2026-01-27
Fix Darwin build and tests; Add MacOS CI

booxter:
This should fix both the build and the tests for Darwin. CI should then help to avoid regressions.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -6,12 +6,32 @@ on: [push, pull_request]
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bvanassche:
The following part of the `.appveyor.yml` file may be helpful:
```
install:
  - cmd: 'ci\openssl.bat'
  - sh:  'sudo ci/install.sh'

build_script:
  - cmd: 'ci\build.bat'
  - sh:  'ci/build.sh'

test_script:
  - cmd: 'ci\test.bat'
  - sh:  'if [ $(uname) != Darwin ]; then ci/test.sh; fi'
  ```

PR #1063, opened 2026-02-13
Block snmp operation till AgentX connection establishes

feroz-ahmed:


feroz-ahmed:
During transient conditions, responding with “No Such Object available on this agent at this OID” gives the false impression that the OID is not implemented. Instead,  returning with “Timeout”, would force Network Management Stations to retry.

PR #1064, opened 2026-02-23
agent: auto_nlist.c: check return value of malloc() calls

rachelrossatx:
## Summary

Three `malloc()` calls in `auto_nlist_value()` lacked NULL checks, causing immediate NULL pointer dereferences if memory allocation fails:

- The `autonlist` node allocation result was passed directly to `memset()` without checking
- The `it->symbol` allocation result was passed directly to `strcpy()` without checking  
- The `it->nl[0].n_name` allocation result was used directly by the AIX/FreeBSD9 code paths without checking

On memory exhaustion, any of these will crash the agent with a NULL pointer dereference rather than failing gracefully.

## Fix

Added NULL checks after each `malloc()` call. On failure, already-allocated memory is freed in reverse order before returning, preventing memory leaks.

## Test plan

- [ ] Build on Linux with address sanitizer and verify no regressions
- [ ] Verify behavior under memory pressure (e.g. with `ulimit -v`)
- [ ] Build on AIX/FreeBSD9 to confirm the `nl[0].n_name` path compiles correctly with the new check

### DIFF
--- a/agent/auto_nlist.c
+++ b/agent/auto_nlist.c
@@ -52,16 +52,33 @@ auto_nlist_value(const char *string)
     }
     if (*ptr == 0) {
         *ptr = (struct autonlist *) malloc(sizeof(struct autonlist));
+        if (*ptr == NULL) {
+            snmp_log(LOG_ERR, "auto_nlist: malloc failed\n");
+            return 0;
+        }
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #1073, opened 2026-03-16
Check if tclist has already been allocated in parser to avoid a memory leak if we re-initialise

eskyuu:
I am trying to use mib_shutdown() followed by mib_init() to reset the loaded MIB tree without launching a new process.  Most things worked, but I discovered a single memory leak where parse.c allocates memry for tclist.

This PR frees the memory used by tclist if it has previously been allocated.

### DIFF
--- a/snmplib/parse.c
+++ b/snmplib/parse.c
@@ -729,6 +729,7 @@ netsnmp_init_mib_internals(void)
 
     memset(nbuckets, 0, sizeof(nbuckets));
     memset(tbuckets, 0, sizeof(tbuckets));
+    free(tclist);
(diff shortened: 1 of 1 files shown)


bvanassche:
> All requested changes have been made.

Thanks!
 
> Out of curiosity - You had asked that I rebase a trivial commit. Do you always want the PR to be presented as a single commit, or for more complex changes would you squash when merging into main?

For every open source project I know, it is preferred that each commit includes a single logical change and also that fixups are squashed into the commit the fixup applies to. Multiple logical changes should be presented as a series of commits with one logical change per commit.

PR #1076, opened 2026-03-20
Sync ifAlias with /sys/class/net/*/ifalias

MichelStam:
This patch adds the ability to Linux to read and write the /sys/class/net/INTERFACE/ifalias when IF-MIB::ifAlias is read or written.

This causes the ifAlias to appear as a description when ip link show is used on the CLI, or in monitoring systems.

Patch was originally written for net-snmp 5.9.1, tested there and then applied on master.

To work the daemon must run as root. This is a limitation of writing / reading /sys which is similar to /proc/sys/net, which is interfaced in ./agent/mibgroup/ip-mib/data_access/scalars_linux.c

The patch has also been tested to return a not writable error when performing SNMP SET while the daemon runs as a low privilege user.


MichelStam:
Changes applied as requested.
Tested again as before (low privilege and as root on 5.9.1). Additional test with a long string.

Cheers, Michel

PR #1083, opened 2026-04-10
if-mib: detect counter discontinuity without missing transition

### LINKED ISSUE #1079: CALCULATE_UCAST unicast counter can decrease or underflow
On Linux where CALCULATE_UCAST computes iucast = iall - imcast (interface.c line 663), the result can decrease or underflow. Team and bond drivers aggregate per-CPU stats non-atomically, so multicast can momentarily exceed total. When multicast is nearly equal to total (for example a NIC in promiscuous mode), the unicast residual oscillates between polls.

Additionally, interface.c has a duplicate unconditional c64_check32_and_update() call for iucast (line 573) that double-processes the counter on non-CALCULATE_UCAST platforms.

Proposed fixes:
1. Clamp CALCULATE_UCAST result: if iucast would decrease while iall is non-decreasing, keep the previous value
2. Remove the duplicate iucast wrap-check call

Found in version 5.9.3.

juhatuFP:
On Linux with bond/team devices, removing or adding slave ports can reset kernel per-CPU counters without the interface leaving the ifTable. ifCounterDiscontinuityTime was only updated on known_missing transitions.

Detect counter decreases (ibytes, iall, obytes) on interfaces that remain present and update ifCounterDiscontinuityTime per RFC 2863 Section 3.1.5.

Fixes #1079.

### DIFF
--- a/agent/mibgroup/if-mib/ifTable/ifTable_data_access.c
+++ b/agent/mibgroup/if-mib/ifTable/ifTable_data_access.c
@@ -394,12 +394,76 @@ _check_interface_entry_for_updates(ifTable_rowreq_ctx *rowreq_ctx,
          * bridge if we come to it).
          */
         if (rowreq_ctx->known_missing) {
+            DEBUGMSGTL(("ifTable:access",
+                     "interface %s returned from missing"
+                     " state, setting discontinuity time\n",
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #1084, opened 2026-04-10
if-mib: fix ifHCInUcastPkts decrease and underflow on bond/team devices

### LINKED ISSUE #1080: ifCounterDiscontinuityTime not updated on counter decrease
When an interface's /proc/net/dev counters decrease (for example due to bond/team reconfiguration or kernel reset), ifCounterDiscontinuityTime
remains at zero. The only code setting it (ifTable_data_access.c line 396) fires on interface name disappearance via the known_missing flag, not on counter decrease. This violates [RFC 2863 3.1.5](https://www.rfc-editor.org/rfc/rfc2863.html#section-3.1.5).

This becomes critical when combined with the CALCULATE_UCAST oscillation (see [1079](https://github.com/net-snmp/net-snmp/issues/1079)). Monitoring systems see a small counter decrease with no discontinuity notification and interpret it as a 64-bit wrap, computing impossible rates.

Proposed fix:
After the known_missing check, compare key counter values against the previous poll. If any decreased, set ifCounterDiscontinuityTime to the current agent uptime.

Found in version 5.9.3.

juhatuFP:
On Linux, ifHCInUcastPkts is computed as iall - imcast (CALCULATE_UCAST). Per-CPU stats aggregation races in bond/team drivers can cause momentary imcast > iall, wrapping u64Subtract to near 2^64, or cause transient iucast decreases.

Save previous iucast/iall before update_stats() and enforce monotonicity in _entry_copy(). When iall is non-decreasing but the computed iucast would decrease (or underflow), hold the previous value.

Remove the duplicate unconditional iucast wrap-check call in update_stats(). When CALCULATE_UCAST is set, new_vals->stats.iucast is always {0,0}, causing a spurious 2^32 adjustment.

Fixes #1080.

### DIFF
--- a/agent/mibgroup/if-mib/data_access/interface.c
+++ b/agent/mibgroup/if-mib/data_access/interface.c
@@ -570,12 +570,14 @@ netsnmp_access_interface_entry_update_stats(netsnmp_interface_entry * prev_vals,
                         "Error expanding ifHCInUcastPkts to 64bits\n"));
         }
 
-        if (0 != netsnmp_c64_check32_and_update(&prev_vals->stats.iucast,
-                                       &new_vals->stats.iucast,
-                                       &prev_vals->old_stats->iucast,
-                                       &need_wrap_check))
-            DEBUGMSGTL(("access:interface",
-                    "Error expanding ifHCInUcastPkts to 64bits\n"));
+        /*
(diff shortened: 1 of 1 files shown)


bvanassche:
Thanks for the patch!

PR #1087, opened 2026-04-22
Limit extension output to 1024 lines to prevent stack-buffer overflow

moshekaplan:
Prevent stack-buffer overflow if output is more than 1024 lines.

### DIFF
--- a/agent/mibgroup/agent/extend.c
+++ b/agent/mibgroup/agent/extend.c
@@ -371,7 +371,7 @@ extend_load_cache(netsnmp_cache *cache, void *magic)
         line_buf[ 0 ] = extension->output;
         for (cp=extension->output; *cp; cp++) {
             if (*cp == '\n') {
-				if (extension->numlines >= 1024) {
+				if (extension->numlines >= (sizeof(line_buf) / sizeof(line_buf[0]))) {
 					break;
 				}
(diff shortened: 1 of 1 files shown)


moshekaplan:
Please note: Truncating the output is only one solution for this problem. Another solution may be preferred.

agent/mibgroup/agent/extend.c
```diff
@@ -371,6 +371,9 @@ extend_load_cache(netsnmp_cache *cache, void *magic)
         line_buf[ 0 ] = extension->output;
         for (cp=extension->output; *cp; cp++) {
             if (*cp == '\n') {
+				if (extension->numlines >= 1024) {
```
bvanassche: Please change `1024` into `sizeof(line_buf) / sizeof(line_buf[0])` or introduce a symbolic name for this constant.

bvanassche (changes requested):

agent/mibgroup/agent/extend.c
moshekaplan: Done

bvanassche:
Please fix your editor settings such that spaces are used for indentation instead of tabs.

The two commits have been combined and have been applied on the master branch.

PR #1088, opened 2026-04-22
Fix comparison logic for interface speed

### LINKED ISSUE #737: interface_openbsd.c: Identical sub-expressions on both sides of operator ">"
Found with SonarCloud

`t_speed > t_speed` will never be true, so this looks like a mistake:

From https://github.com/net-snmp/net-snmp/blob/5b74f2738187906e8000d6f8d8411861669c9607/agent/mibgroup/if-mib/data_access/interface_openbsd.c#L374:
```
                if (t_speed_high > m_speed_high ||
                    (t_speed_high == m_speed_high && t_speed > t_speed)) {
```

moshekaplan:
Fixes https://github.com/net-snmp/net-snmp/issues/737


bvanassche:
Thanks for the patch!

PR #1093, opened 2026-05-12
Fix stack buffer overflows in pass_common SET and hex formatter

TristanInSec:
## Summary

Fix two stack buffer overflows in the pass/passthrough MIB handler:

1. **`pass_common.c` SET handler** (`netsnmp_internal_pass_set_format`): A `memcpy` at line 253 copies `var_val_len` bytes into a stack buffer `buf2` without checking the length. An SNMP SET with an OCTET STRING value larger than `sizeof(buf2)` overflows the buffer. Add a length cap before the copy.

2. **`pass_common.c` hex formatter** (`netsnmp_internal_bin2asc`): The hex expansion loop writes 3 bytes per input byte back into the caller's buffer. When the input length exceeds `SNMP_MAXBUF / 3`, the formatted output overflows the buffer. Cap the input length so the output fits.

## Testing

Built and verified the fixes compile cleanly. Confirmed the overflow paths are reachable via standard SNMP SET operations.

### DIFF
--- a/agent/mibgroup/ucd-snmp/pass_common.c
+++ b/agent/mibgroup/ucd-snmp/pass_common.c
@@ -38,9 +38,9 @@ netsnmp_internal_bin2asc(char *p, size_t n)
     int             i, flag = 0;
     char            buffer[SNMP_MAXBUF];
 
-    /* prevent buffer overflow */
-    if ((int)n > (sizeof(buffer) - 1))
-        n = sizeof(buffer) - 1;
+    /* prevent buffer overflow: hex path writes 3 bytes per input byte */
+    if (n > sizeof(buffer) / 3 - 1)
+        n = sizeof(buffer) / 3 - 1;
 
     for (i = 0; i < (int) n; i++) {
         buffer[i] = p[i];
@@ -245,6 +245,8 @@ netsnmp_internal_pass_set_format(char *buf,
                 (int) ((utmp & 0xff)));
         break;
     case ASN_OCTET_STR:
+        if (var_val_len > sizeof(buf2))
+            var_val_len = sizeof(buf2);
         memcpy(buf2, var_val, var_val_len);
         if (var_val_len == 0)
             sprintf(buf, "string \"\"");


agent/mibgroup/ucd-snmp/pass_common.c
```diff
@@ -51,6 +51,12 @@ netsnmp_internal_bin2asc(char *p, size_t n)
         p[n] = 0;
         return n;
     }
+    /*
+     * The hex expansion writes 3 bytes per input byte back into p.
+     * Cap n so the output fits in the caller's SNMP_MAXBUF-sized buffer.
+     */
+    if (n > (sizeof(buffer) / 3))
+        n = sizeof(buffer) / 3;
```
bvanassche: Is this new check correct? Shouldn't it be changed into the following to take the terminating NUL into account?
```
if (n > sizeof(buffer) / 3 - 1)
    n = sizeof(buffer) / 3 - 1
```
Instead of adding this check here, replace the check under "prevent buffer overflow" with an improved check.

bvanassche (changes requested):

agent/mibgroup/ucd-snmp/pass_common.c
TristanInSec: Good catch. Consolidated into the existing guard with `sizeof(buffer) / 3 - 1` and dropped the separate hex-path check. Updated in e766381.

bvanassche:
Please combine the two commits into a single commit, e.g. with `git rebase -i`.

bvanassche:
The reported buffer overflows have been fixed in a different way. Please take a look at commits 2c16eea446efdd94fad789a2fea2f1c3620a611b and cb91da8a2cfb312449b539b00b49cd16bfa64160.

TristanInSec:
> The reported buffer overflows have been fixed in a different way. Please take a look at commits [2c16eea](https://github.com/net-snmp/net-snmp/commit/2c16eea446efdd94fad789a2fea2f1c3620a611b) and [cb91da8](https://github.com/net-snmp/net-snmp/commit/cb91da8a2cfb312449b539b00b49cd16bfa64160).

Hi Bart,

Thanks for fixing these. I reviewed both commits (2c16eea and cb91da8) and the refactored APIs look solid. The length-aware signatures are a better approach than the bounds checks I proposed.

Two requests if you don't mind:

1. Could you open a GitHub Security Advisory (GHSA) for the 2 buffer overflows? This would get a CVE assigned through the GitHub CNA and provide a formal tracking reference for downstream distros. Happy to be added as a collaborator/credit.

2. Would you be willing to add a Reported-by trailer to a follow-up commit? Something like:
   Reported-by: Tristan Madani

Thanks again for the quick turnaround on the fixes.

Best,
Tristan

bvanassche:
>   Could you open a GitHub Security Advisory (GHSA) for the 2 buffer overflows? This would get a CVE assigned through the GitHub CNA and provide a formal tracking reference for downstream distros. Happy to be added as a collaborator/credit.

Sorry but I'm not familiar with this process ...

>  Would you be willing to add a Reported-by trailer to a follow-up commit? Something like:
    Reported-by: Tristan Madani

Right, I should have mentioned your name in the fixes. I'm not sure how to do this now without rewriting git history since several other commits have already been applied on top of the commits mentioned above?

TristanInSec:
> > Could you open a GitHub Security Advisory (GHSA) for the 2 buffer overflows? This would get a CVE assigned through the GitHub CNA and provide a formal tracking reference for downstream distros. Happy to be added as a collaborator/credit.
> 
> Sorry but I'm not familiar with this process ...
> 
> > Would you be willing to add a Reported-by trailer to a follow-up commit? Something like:
> > Reported-by: Tristan Madani
> 
> Right, I should have mentioned your name in the fixes. I'm not sure how to do this now without rewriting git history since several other commits have already been applied on top of the commits mentioned above?

Hi Bart,

No issues at all with the alternative fix, thanks for taking care of it.

On the GHSA: if you enable GitHub Security Advisories on the repo (Settings > Security > Advisories), I can draft the advisory and add you as a collaborator. You would just need to review and publish it. GitHub's CNA handles CVE assignment automatically from there.

On credit: no need to rewrite history. A follow-up empty commit works fine:
git commit --allow-empty -m "Credit Tristan Madani for reporting the pass_common buffer overflows fixed in 2c16eea and cb91da8"

Thanks

hardaker:
Hi @TristanInSec ,

Thanks for pointing out the issue.  We do not generally create CVEs for portions of the code that are past authentication and authorization and for users that already have write access to the MIB tree, which this issue falls into.  The number of actual people impacted by this issue is likely close to zero, though it's still an excellent bug to fix.

TristanInSec:
> Hi @TristanInSec ,
> 
> Thanks for pointing out the issue. We do not generally create CVEs for portions of the code that are past authentication and authorization and for users that already have write access to the MIB tree, which this issue falls into. The number of actual people impacted by this issue is likely close to zero, though it's still an excellent bug to fix.

Hi Wes,

Thanks for the feedback. I understand the position on post-auth issues, but net-snmp does have precedent: CVE-2022-24807 (SET to vacmAccessTable causing out-of-bounds memory access) and CVE-2022-24806 (SET with malformed OIDs) both required read-write credentials and were assigned CVEs. The pass_common overflows are the same class -- authenticated SET handling with stack buffer overflows. And in SNMPv1/v2c deployments, which net-snmp still widely supports, "authentication" is a plaintext community string often left at defaults like "private".

hardaker:
The huge difference between the ones you point out and the one you've identified is the liklihood of exploitation.  The OID one is the largest and is pre-authentication and authorization, and the vacm table is a very common table (though write access to that table is typically limited to admins anyway (or at least should be).  For pass_common it's far less used, likely only by administrators with read-write shell access too, and you'd have to know how it was deployed on the agent in question as its an extension.  So the likelihood of real-world exploit is very very low.

I can create a CVE eventually but other timing constraints are in the way and it'll be a week or two probably.

hardaker:
Hi, one more thing for the credit: what AI agent did you use in the process of finding the bug?

TristanInSec:
> I can create a CVE eventually but other timing constraints are in the way and it'll be a week or two probably.

That would be much appreciated! 
I can also request one with a CNA I collaborate with if that's easier, happy to do so.

bvanassche:
On 7/6/26 10:27 AM, Tristan wrote:
> On credit: no need to rewrite history. A follow-up empty commit
> works fine:
> 
> git commit --allow-empty -m "Credit Tristan Madani for reporting
> the pass_common buffer overflows fixed in 2c16eea <https://github.com/ 
> net-snmp/net-snmp/commit/2c16eea446efdd94fad789a2fea2f1c3620a611b> and 
> cb91da8 <https://github.com/net-snmp/net-snmp/commit/ 
> cb91da8a2cfb312449b539b00b49cd16bfa64160>"

Please take a look at commit db43905e8c09f31514195422b95f8f063099f388.

TristanInSec:
> On 7/6/26 10:27 AM, Tristan wrote: On credit: no need to rewrite history. A follow-up empty commit works fine: git commit --allow-empty -m "Credit Tristan Madani for reporting the pass_common buffer overflows fixed in [2c16eea](https://github.com/net-snmp/net-snmp/commit/2c16eea446efdd94fad789a2fea2f1c3620a611b) <https://github.com/ net-snmp/net-snmp/commit/2c16eea446efdd94fad789a2fea2f1c3620a611b> and [cb91da8](https://github.com/net-snmp/net-snmp/commit/cb91da8a2cfb312449b539b00b49cd16bfa64160) <https://github.com/net-snmp/net-snmp/commit/ [cb91da8](https://github.com/net-snmp/net-snmp/commit/cb91da8a2cfb312449b539b00b49cd16bfa64160)>"
> Please take a look at commit [db43905](https://github.com/net-snmp/net-snmp/commit/db43905e8c09f31514195422b95f8f063099f388).

Thanks Bart, much appreciated!

TristanInSec:
> I can create a CVE eventually but other timing constraints are in the way and it'll be a week or two probably.

Hi @hardaker, just to follow up on this point, I offer to handle this part to save you some admin work. Sounds good to you? Thanks.

hardaker:
Hi @TristanInSec ...  I was waiting for you to answer my question: "Hi, one more thing for the credit: what AI agent did you use in the process of finding the bug?"

TristanInSec:
> Hi @TristanInSec ... I was waiting for you to answer my question: "Hi, one more thing for the credit: what AI agent did you use in the process of finding the bug?"

Hi @hardaker,
To answer your question: I use custom SAST scanners I created (closed source) for initial pattern detection across codebases, then manually triage and trace each finding to confirm exploitability before reporting.
Thanks.

hardaker:
Excellent information @TristanInSec thank you.  Is your custom scanner based on top of an AI agent/etc and if so which one and what version?

hardaker:
CVE request created, assignment TBD

PR #1094, opened 2026-05-14
bug: disman-ping: Fail to set pingCtlProbeCount, and missmatch of packet PID and system PID

Torstein-Eide:
## Background

`T154dismanpingmib_simple` creates DISMAN-PING-MIB control rows, configures them for IPv4 loopback probes, activates the rows, and checks that `pingResultsSentProbes` and `pingResultsProbeResponses` both reach the configured probe count.

## Issue

`pingCtlProbeCount` had a write handler but was registered as read-only, causing the multi-varbind setup SET to fail before pingCtlAdminStatus could be enabled.

After that was fixed, ping replies were still ignored when the agent PID exceeded 16 bits. ICMP echo packets only carry a 16-bit identifier, while system PIDs can be wider, so comparing the packet identifier against the full PID caused valid replies to be discarded.

## Fix

Register `pingCtlProbeCount` as writable so row setup can complete.

Mask the PID to 16 bits before using it as the IPv4 ICMP echo identifier, matching the protocol field size.

Add comments to the simple DISMAN ping test describing the tested OIDs and test phases.


bvanassche:
Thanks for the patch!

PR #1097, opened 2026-05-27
ifXTable: fix null-terminate ifName string in ifName_get in ifXTable.c

AntonMoryakov:
Static analyzer detected string copying without a null terminator in the ifName_get function (line 438).

Issue:
- strlen() returns length excluding the '\0' terminator
- exactly tmp_len bytes were allocated and copied
- resulting buffer was not null-terminated → undefined behavior

Fix:
- allocate tmp_len + 1 bytes via malloc()
- copy tmp_len + 1 bytes via memcpy() to include the '\0' terminator
- *ifName_val_ptr_len_ptr remains tmp_len (per SNMP DisplayString requirements)

Fixes static analyzer warning: 'Copying from string without null termination' The MFD helper will automatically free the allocated memory.


bvanassche:
Thanks for the patch!

PR #1104, opened 2026-06-19
OSS-Fuzz: Fix build error for fuzzer

arthurscchan:
OSS-Fuzz only allows logging into /tmp directory and this PR fixes the fuzzer build error to ensure the logging path is configured to /tmp directory.


bvanassche:
On my test setup the Net-SNMP fuzz tests build fine with and without this patch. So it's not clear to me why the title of this pull request says "fix build error". If this pull request really fixes a build error, please copy/paste it into the patch description.

arthurscchan:
Apology for the confusion and rush on this. I have retest the whole thing and found that the previous crash in build is because of some random timeout issue. Sincere apology again for wasting your time on this.

PR #1114, opened 2026-07-05
Fix UAF in expression function evaluation

Jack-Kaplan:
Fixes Coverity 454601.
`_expValue_evalFunction()` frees `params`, which is `func->next_variable`, but leaves `func->next_variable` pointing at the freed node. The caller immediately reads that pointer via `vp1 = vp->next_variable`.
Relink `func` to `params->next_variable` before detaching and freeing `params`, preserving the apparent linked-list intent and avoiding the dangling pointer.


Jack-Kaplan:
I was trying to understand the intent behind this. It seemed to me to be a linked-list. If this understanding is wrong and this is not the correct fix, I apologize for the incorrect fix.

bvanassche:
Thanks for the patch!

PR #1116, opened 2026-07-16
Remove duplicate entry from contributors list

moshekaplan:
"Marus Meissner" in the Thanks list should be "Marcus Meissner" (confirmed by a later duplicate entry "Marcus Meissner" with the same email, so this is clearly a typo/duplicate).


bvanassche:
Thanks for the patch!

PR #1117, opened 2026-07-20
snmpTlstmAddrTable.c - Avoid potential NULL deref on addrs_itr

moshekaplan:
Fix coverity 85485 - only release iterator if it is set


moshekaplan:
Please note: Although this fix seems to make sense to me, I do not have the ability to test it.

PR #1118, opened 2026-07-20
client_intf.c - Avoid potential null deref

moshekaplan:
Fixes coverity 379734

### DIFF
--- a/python/netsnmp/client_intf.c
+++ b/python/netsnmp/client_intf.c
@@ -1384,8 +1384,6 @@ netsnmp_get_or_getnext(PyObject *self, PyObject *args, int pdu_type,
   size_t oid_arr_len = MAX_OID_LEN;
   int type;
   char *str_buf = NULL;
-  const char *tag;
-  const char *iid;
   int getlabel_flag = NO_FLAGS;
   int sprintval_flag = USE_BASIC;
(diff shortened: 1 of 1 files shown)


moshekaplan:
Please note: Although this fix seems to make sense to me, I do not have the ability to test it.

python/netsnmp/client_intf.c
```diff
@@ -1384,8 +1384,8 @@ netsnmp_get_or_getnext(PyObject *self, PyObject *args, int pdu_type,
   size_t oid_arr_len = MAX_OID_LEN;
   int type;
   char *str_buf = NULL;
-  const char *tag;
-  const char *iid;
+  const char *tag = NULL;
+  const char *iid = NULL;
```
bvanassche: The description of this commit is too cryptic. Please explain in the commit message why this change has been made.

Additionally, please move the declaration of both variables inside the only while-loop in which these are used.

bvanassche (changes requested):

python/netsnmp/client_intf.c
moshekaplan: Sorry about that. Should be better now. Can you take a look?

bvanassche:
Thanks for the patch!

PR #690, opened 2023-09-11
V5 9 patches

nielsb:
Prepare for FreeBSD 15.0
Restore the correct NEWS/CHANGES for 5.9.2

Issue #640, opened 2023-08-27
Fix build on FreeBSD 15


jurajlutter:
FreeBSD main branch is now 15.0-CURRENT.

Please see https://reviews.freebsd.org/D41611 for a fix to build on FreeBSD 15.0.

bvanassche:
How to test these patches? I don't see FreeBSD 15 being mentioned on https://www.freebsd.org/where/.

jurajlutter:
Unfortunately, there ISO or VM images for 15.0-CURRENT are not yet available (as 15.0-CURRENT was introduced in git last thursday).

I can provide an access to 15.0-CURRENT VM, if needed.

bvanassche:
It would be great if access to a FreeBSD 15 VM could be provided.

nielsb:
I have just verified building on FreeBSD CURRENT 20230907 with a fixed freebsd15.h
See PR #690 
You get download an image from download.freebsd.org/snapshots/

jeneric:
Building on FreeBSD 15 works since #690, but that commit added
`include/net-snmp/system/freebsd15.h` without adding it to `Makefile.in`'s
`INCLUDESUBDIRHEADERS`, so the header is never installed. The installed
`net-snmp-config.h` still says

```c
#define NETSNMP_SYSTEM_INCLUDE_FILE "net-snmp/system/freebsd15.h"
```

which means net-snmp itself builds, but anything compiling against the installed
headers fails on the missing include. `freebsd16.h` has the same problem.

I have opened #1121 to add both to the install list. Field evidence in the PR:
with the equivalent patch applied, a downstream distribution that builds
net-snmp from source now passes on FreeBSD 15.0-RELEASE-p5, 15.1-RELEASE-p1 and
16.0-CURRENT, where the release before it had 28 FreeBSD reports and no passes.

Flagging it here since this is the FreeBSD 15 thread, though it is a separate
defect from the compile failure this issue was opened for.

PR #1121, opened 2026-07-31
Makefile.in: install the FreeBSD 15 and 16 system headers

jeneric:
`include/net-snmp/system/freebsd15.h` and `freebsd16.h` are in the source tree
but are not listed in `Makefile.in`'s `INCLUDESUBDIRHEADERS`, which stops at
`freebsd14.h`. They are therefore never installed, while the installed
`net-snmp-config.h` still contains:

```c
#define NETSNMP_SYSTEM_INCLUDE_FILE "net-snmp/system/freebsd15.h"
```

so every compile against the installed headers fails with a missing include.

### How the gap appeared

`freebsd15.h` arrived in 2230933c7 ("Prepare for FreeBSD 15", 2023-09-11) and
`freebsd16.h` in 106cac2f0 (2025-09-11). Neither commit touched `Makefile.in`.

The convention elsewhere is to do both: 7ffa1ced1 ("support openbsd8",
2025-03-08) added `openbsd8.h` and the matching `INCLUDESUBDIRHEADERS` entry in
the same commit, which is why `openbsd8.h` installs correctly today.

This is also not dormant. 44f905318 ("configure: refine freebsd header fallback
to dynamically pick the highest shipped freebsdNN.h", 2026-03-09) made configure
resolve a FreeBSD 17 host to `freebsd16`, so the selection logic actively reaches
for a header that the install list does not carry.

### Verification

The equivalent change has been run on FreeBSD, from two independent directions.

**CI.** The `Alien::SNMP` distribution builds net-snmp 5.9.5.2 from source and
has been applying this fix at build time. Its GitHub Actions matrix runs FreeBSD
13.5, 14.4 and 15.1, and the full matrix on its master branch is green on all
three, 20 jobs of 20. On 15.1 the build log shows the two headers being added to
the install list, and the suite then compiles and links against the installed
headers, which is precisely what fails without this change.

**CPAN Testers.** The release carrying that fix has 10 passing FreeBSD reports
and no failures: seven on 15.0-RELEASE-p5, two on 15.1-RELEASE-p1 and one on
**16.0-CURRENT**, spanning perl 5.14 through 5.45. The previous release, without
it, had 28 FreeBSD reports and no passes.

Two notes on what that does and does not prove. `Alien::SNMP` computes the
missing names at build time rather than hardcoding them, so what ran on FreeBSD
is the same two names reaching the same `INCLUDESUBDIRHEADERS` variable and the
same install rule, against 5.9.5.2 rather than against this branch. And the
static property is checked here directly: after this change every `.h` in
`include/net-snmp/system/` appears in `INCLUDESUBDIRHEADERS`, and every name in
`INCLUDESUBDIRHEADERS` exists in that directory.

### Related

Not a fix for #640, which was about compiling on FreeBSD 15 and was answered by
#690. This is the follow-on gap that #690 left: the header was added, the
install list was not updated.

The same two headers are missing from `INCLUDESUBDIRHEADERS` on `V5-9-patches`.
Happy to open a backport if that is wanted.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -21,7 +21,7 @@ INCLUDESUBDIRHEADERS= aix.h bsd.h bsdi3.h bsdi4.h bsdi.h cygwin.h \
 	darwin.h dragonfly.h dynix.h \
 	freebsd2.h freebsd3.h freebsd4.h freebsd5.h freebsd6.h \
 	freebsd7.h freebsd8.h freebsd9.h freebsd10.h freebsd11.h \
-	freebsd12.h freebsd13.h freebsd14.h freebsd.h \
+	freebsd12.h freebsd13.h freebsd14.h freebsd15.h freebsd16.h freebsd.h \
 	generic.h \
 	hpux.h irix.h kfreebsd.h linux.h mingw32.h mingw32msvc.h mips.h \
 	netbsd.h nto-qnx6.h osf5.h \


bvanassche:
Thanks for the patch!

PR #1124, opened 2026-08-14
Fix commit ab94dd80 (Windows: Fix 64-bit compatibility issues)

nielsb:
when on a non-Windows system


bvanassche:
Thanks for the patch!

On which operating system did the build fail? It's unfortunate that this was not caught by any of the existing CI tests.

nielsb:
It was Linux in general, Alma (RHEL), Fedora, Ubuntu, as soon as you select SSH transport

PR #1127, opened 2026-08-30
Don't spam default domain

dwhitemv25:
Specifying default domains here spams user-configured default domains set via `defDomain` in snmp.conf, rendering the `defDomain` keyword useless for the shipped snmp* tools.

For instance, this fix allows specifying `defDomain snmp udp6 udp` in snmp.conf to set a preference for IPv6 addresses in snmpget, et. al.

The default is still `udp udp6`, as set in _init_snmp, so there is no user visible change out-of-box.

The tcp case is used only by agentx so it does not need the same fix.


bvanassche:
Thanks for the patch!

PR #1137, opened 2026-09-21
GHSL-2026-165 libsnmp: Derive TLSTM securityName from the verified certificate chain

Kwstubbs:
netsnmp_openssl_get_cert_chain() built its candidate list from SSL_get_peer_cert_chain(), which reports the certificates exactly as the peer sent them rather than the chain that OpenSSL verified. Every entry was then handed to netsnmp_cert_get_secname_maps() for fingerprint matching, so a certificate that took no part in path validation could still select an snmpTlstmCertToTSNTable row and decide the tmSecurityName for the session.

The snmpTlstmCertToTSNTable DESCRIPTION only considers a row a match when its fingerprint identifies the presented certificate, or a locally held copy of a trusted CA certificate that was used to validate the path to the presented certificate. Walk SSL_get0_verified_chain() instead and keep only those entries that are CA certificates in the local store, which leaves both documented cases working while dropping certificates that are merely present in the peer's Certificate message.

verify_callback() deliberately accepts a locally known peer certificate even when its issuer cannot be checked, and OpenSSL notes that the chain reported for such a session may be incomplete. Since that chain is not a validated path, consult it only when SSL_get_verify_result() reports X509_V_OK; otherwise just the presented certificate is used, which is the only row the table permits in that case.

Add a configure check for SSL_get0_verified_chain(); when it is not available, only the presented certificate is used for row lookups.


bvanassche:
Thanks for the patch!

PR #1138, opened 2026-09-21
GHSL-2026-164 libsnmp: Keep the Unix domain transport data type consistent

Kwstubbs:
netsnmp_unix_accept() stored a bare struct sockaddr_un in t->data while every other function in this file reads that member as a sockaddr_un_pair, so the accepted peer address was reinterpreted through a larger and differently shaped type.

Allocate a sockaddr_un_pair and keep the accepted peer address in its client member, and set data_length accordingly so that the copy made by netsnmp_transport_copy() stays correctly sized.

While here, only let netsnmp_unix_close() remove the socket path the transport created itself. The address of the far end belongs to the peer, so it is not ours to unlink. That branch had no effect before, since the client member was never filled in.


bvanassche:
Thanks for the patch!

